PR #14, opened 2026-07-01
Add in-process Rust SDK

ZhiXiao-Lin:
## Summary
- add a workspace-local ctx-sdk crate with an in-process Rust facade over ctx history primitives
- expose typed APIs for status/init, source discovery, provider import, search, session/event lookup, and doctor checks
- add behavior-oriented SDK unit/integration coverage using real local-history formats, plus an opt-in live LLM test
- document SDK usage and add the Rust SDK page to embedded CLI docs

## Notes
- this intentionally opens the SQLite store and calls ctx-history-* crates directly; it does not shell out to the ctx CLI or parse CLI JSON
- the crate is publish=false for now because the lower-level ctx-history-* crates are still internal workspace crates
- the live LLM test is ignored and env-gated so default CI does not depend on network, API keys, provider quotas, or model availability
- live LLM smoke supports OpenAI Responses and OpenAI-compatible Chat Completions providers via CTX_SDK_LIVE_OPENAI_API=chat
- tests avoid fixture-specific IDs and exact internal counters where possible; they assert public SDK contracts such as discovery, import success, searchability, citations, and typed transcript lookup
- private API keys, private base URLs, private model names, and local agent configuration files are not committed; live smoke values are supplied only through environment variables
- the live test passes the authorization header to curl through stdin config, so the API key is not exposed in process arguments

## Coverage
- cargo llvm-cov -p ctx-sdk --summary-only
- line coverage: 92.92%
- region coverage: 91.33%
- function coverage: 96.10%
- production module line coverage: client 93.16%, import 87.16%, search 85.71%, transcript 96.00%

## Tests
- cargo fmt --all --check
- cargo test -p ctx-sdk
- cargo llvm-cov -p ctx-sdk --summary-only
- cargo check -p ctx
- cargo test --workspace
- live OpenAI-compatible Chat Completions smoke was run with private endpoint, model, and credentials exported in the local environment; those values are not included in this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,6 +356,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctx-sdk"
+version = "0.11.0"
+dependencies = [
+ "chrono",
+ "ctx-history-capture",
+ "ctx-history-core",
+ "ctx-history-search",
+ "ctx-history-store",
+ "rusqlite",
+ "serde_json",
+ "tempfile",
+ "thiserror",
+ "uuid",
+]
+
+[[package]]
 name = "difflib"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "crates/ctx-cli",
+    "crates/ctx-sdk",
     "crates/ctx-history-capture",
     "crates/ctx-history-core",
     "crates/ctx-history-search",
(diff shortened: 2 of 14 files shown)


luca-ctx:
Thanks for the idea and the PR!

We took a step back and approached this from a broader sense in that we want to support many language SDKs, including rust.

We have merged it in now to main, please take a look and let me know if this will work for you.

Since we are early in our rollout of this, we are going to hold off on publishing to third party package manager artifact stores like cargo until everything is confirmed to be stabilized as far as the API, so we don't have to do a lot of backwards compatibility or migration work.

For now, you can adopt by using a git hash / tag to point to the specific version.

I will close this for now as it is superseded by the work on main. Thank you for the contribution!

PR #15, opened 2026-07-01
fix(store): reset stale index metadata when cataloged files change

atharva-patill:
## Summary

When a cataloged transcript changes, `indexed_status` is reset to `pending`, but the remaining `indexed_*` metadata continues to retain its previous values because the corresponding `CASE` expressions return the existing value in both branches.

This updates those fields to reset to `NULL` when the file changes, avoiding stale indexing metadata and bringing the behavior in line with the existing `source_import_files` UPSERT and `invalidate_provider_import_indexes()`.

## Validation

- Updated the existing `catalog_upsert_preserves_index_state_until_file_changes` test
- All tests pass

### DIFF
--- a/crates/ctx-history-store/src/lib.rs
+++ b/crates/ctx-history-store/src/lib.rs
@@ -1502,19 +1502,19 @@ impl Store {
                     WHEN catalog_sessions.file_size_bytes = excluded.file_size_bytes
                      AND catalog_sessions.file_modified_at_ms = excluded.file_modified_at_ms
                     THEN catalog_sessions.indexed_at_ms
-                    ELSE catalog_sessions.indexed_at_ms
+                    ELSE NULL
                 END,
                 indexed_file_size_bytes = CASE
                     WHEN catalog_sessions.file_size_bytes = excluded.file_size_bytes
                      AND catalog_sessions.file_modified_at_ms = excluded.file_modified_at_ms
                     THEN catalog_sessions.indexed_file_size_bytes
-                    ELSE catalog_sessions.indexed_file_size_bytes
+                    ELSE NULL
                 END,
                 indexed_file_modified_at_ms = CASE
                     WHEN catalog_sessions.file_size_bytes = excluded.file_size_bytes
                      AND catalog_sessions.file_modified_at_ms = excluded.file_modified_at_ms
                     THEN catalog_sessions.indexed_file_modified_at_ms
-                    ELSE catalog_sessions.indexed_file_modified_at_ms
+                    ELSE NULL
                 END,
                 indexed_status = CASE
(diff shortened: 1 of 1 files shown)


luca-ctx:
Thanks again for identifying this. The stale `indexed_*` metadata was a bug: once a cataloged transcript changed, the row could be pending for the current file while still showing completion metadata from the previous version.

I merged the fix in PR #17 and tracked it in issue #16. The main difference from this patch is that ctx also used the old indexed byte size as the Codex append-only tail checkpoint. Clearing all `indexed_*` fields on change fixed the misleading metadata, but it also made append refresh reprocess old JSONL events.

The merged fix separates those concepts: `indexed_*` now only describes whether the current cataloged file version is fully indexed, and new `last_imported_*` checkpoint fields preserve safe incremental tail refresh state. Changed files clear stale completion metadata, append-only Codex growth can still import just the appended tail, and shrink/same-size rewrite cases do not keep trusting the old checkpoint.

Thanks so much for finding the bug and proposing the first version of the fix. It will be incorporated in the upcoming 0.13 release. Let me know if this solves it for you or if you have any other questions!

atharva-patill:
> Thanks again for identifying this. The stale `indexed_*` metadata was a bug: once a cataloged transcript changed, the row could be pending for the current file while still showing completion metadata from the previous version.
> 
> 
> 
> I merged the fix in PR #17 and tracked it in issue #16. The main difference from this patch is that ctx also used the old indexed byte size as the Codex append-only tail checkpoint. Clearing all `indexed_*` fields on change fixed the misleading metadata, but it also made append refresh reprocess old JSONL events.
> 
> 
> 
> The merged fix separates those concepts: `indexed_*` now only describes whether the current cataloged file version is fully indexed, and new `last_imported_*` checkpoint fields preserve safe incremental tail refresh state. Changed files clear stale completion metadata, append-only Codex growth can still import just the appended tail, and shrink/same-size rewrite cases do not keep trusting the old checkpoint.
> 
> 
> 
> Thanks so much for finding the bug and proposing the first version of the fix. It will be incorporated in the upcoming 0.13 release. Let me know if this solves it for you or if you have any other questions!

Thanks @luca-ctx so much for the detailed explanation! That makes a lot of sense. I hadn’t realized the indexed byte size was also being used as the append checkpoint, so I can see why simply clearing those fields would’ve caused problems. Splitting the indexing metadata from the incremental import state feels like a much cleaner solution.

I really appreciate you taking the time to explain the reasoning behind the change, and thanks for the kind words as well. I’m glad I could help, even if it was just by uncovering the issue. I’ll keep exploring the new codebase and hopefully contribute a few more fixes along the way.

PR #23, opened 2026-07-02
fix(cli): improve error for nonexistent import paths

atharva-patill:
## Summary

When `ctx import --path` is given a path that doesn't exist, the command currently falls through to the filesystem scan and returns a generic OS error.

This adds an early check for explicit import paths and returns a clearer error message instead. It also adds an integration test to cover this behavior.

### Before

Screenshot 2026-07-02 at 10 28 58 AM

### After

Screenshot 2026-07-02 at 12 29 00 PM

## Validation

- Added an integration test for `ctx import --path` with a nonexistent path
- New test passes


luca-ctx:
@Atharvapatil2005 Thanks so much for the bugfix!

atharva-patill:
> @Atharvapatil2005 Thanks so much for the bugfix!

@luca-ctx You're Welcome! Glad I could help.

PR #38, opened 2026-07-03
fix(cli): prevent overflow panic in --since parsing

atharva-patill:
## Summary

While testing `--since`, I noticed that passing a very large day value (for example `500000000d`) causes the CLI to panic because of unchecked date arithmetic.

This changes that behavior to return a normal user-facing error instead of panicking. I've also added a regression test to cover this case.

### Before

```console
$ ctx search foo --since 500000000d

thread 'main' panicked at crates/ctx-cli/src/main.rs:6417:19:
DateTime - TimeDelta overflowed
```

### After

```console
$ ctx search foo --since 500000000d

Error: invalid --since day window: 500000000d: value too large
```

## Validation

- Added a regression test for oversized `--since` values
- Relevant unit tests pass

### DIFF
--- a/crates/ctx-cli/src/main.rs
+++ b/crates/ctx-cli/src/main.rs
@@ -6414,7 +6414,12 @@ fn parse_since_filter(value: &str) -> Result<chrono::DateTime<Utc>> {
         let days: i64 = days
             .parse()
             .with_context(|| format!("invalid --since day window: {value}"))?;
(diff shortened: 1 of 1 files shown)


luca-ctx:
Thank you for the bugfix!

atharva-patill:
> Thank you for the bugfix!

@luca-ctx You're welcome!

PR #46, opened 2026-07-04
Fall back to USERPROFILE when resolving home for provider discovery

sin5ddd:
## Summary

- Fix provider history discovery finding no sources on Windows: the CLI resolved the home directory from `HOME` only, which is normally unset there, so `ctx import --all`, `ctx sources`, and the search auto-refresh all came back empty.
- Reuse the existing resolution chain from `identity.rs` (`HOME` → `USERPROFILE` → `HOMEDRIVE`+`HOMEPATH`) instead of duplicating it or pulling in a new dependency.
- Add two regression tests that run discovery with `HOME` unset and only `USERPROFILE` set; they are env-driven, so Linux CI exercises the Windows path too.

## Other Changes

- Drop provider override variables inherited from the host machine (`HERMES_HOME`, `SHELLEY_DB`, `OPENCLAW_STATE_DIR`, `ASTRBOT_ROOT`) in the shared test helpers, so import tests never discover the developer's real agent history.

## Checks

- `cargo fmt --all`
- `cargo clippy --workspace --all-targets`
- `cargo test -p ctx --test cli`
- Verified on Windows 11: `ctx sources` discovers claude/codex/cursor/opencode, `ctx import --all` imports 273 sessions from 6 sources, and search auto-refresh completes instead of silently skipping.

### DIFF
--- a/crates/ctx-cli/src/identity.rs
+++ b/crates/ctx-cli/src/identity.rs
@@ -134,7 +134,9 @@ fn non_empty_env_path(key: &str) -> Option<PathBuf> {
         .map(PathBuf::from)
 }
 
-fn home_dir() -> Option<PathBuf> {
+/// Resolve the user home directory from `HOME`, falling back to the
+/// Windows `USERPROFILE` and `HOMEDRIVE`+`HOMEPATH` conventions.
+pub(crate) fn home_dir() -> Option<PathBuf> {
     non_empty_env_path("HOME")
         .or_else(|| non_empty_env_path("USERPROFILE"))
         .or_else(|| {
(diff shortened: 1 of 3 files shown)


atharva-patill:
Hey @sin5ddd, quick question—on Windows, is HOME typically unset by default? I hadn’t realized that until I read through this PR.

sin5ddd:
Yes, By default on Windows, `%USERPROFILE%` is used as the environment variable for the home directory, while `HOME` remains unset unless configured manually.
While `HOME` is available in environments like Git Bash and WSL, it is not available in native PowerShell.

Issue #40, opened 2026-07-03
pi source tries ~/.pi/sessions.jsonl which shouldn't even exist


ahirner:
`ctx` tries to find and process such file, but that is never produced.
```console
$ ctx sources
codex /Users/abcd/.codex/sessions available (codex_session_jsonl_tree)
codex /Users/abcd/.codex/history.jsonl available (codex_history_jsonl)
pi /Users/abcd/.pi/sessions.jsonl missing (pi_session_jsonl)
...
```

I quickly grepped through the pi histry. It seems they worked with session directory ever since, e.g. https://github.com/earendil-works/pi/pull/2598. One PR tries to migrate to sqlite FWIW https://github.com/earendil-works/pi/pull/6227

It seems I'm missing something obvious.

PR #50, opened 2026-07-04
Add OMP (~/.omp/agent/sessions) to Pi provider default locations

msokh:
## Motivation

[OMP (Oh My Pi)](https://github.com/can1357/oh-my-pi) is a Pi fork that renames the config root from `.pi` to `.omp`. Session history lives at `~/.omp/agent/sessions` in the identical per-session JSONL format Pi already uses.

Source (upstream OMP): `packages/utils/src/dirs.ts` — `CONFIG_DIR_NAME = ".omp"`, and `getSessionsDir()` resolves to `~/.omp/agent/sessions`.

Without this change, `ctx sources` / `ctx import --all` never discover OMP history natively; only an explicit `ctx import --provider pi --path ~/.omp/agent/sessions` works.

## Approach

Adds a second `ProviderDefaultLocation` entry to `PI_DEFAULTS` (`crates/ctx-history-capture/src/provider_sources.rs`) for `~/.omp/agent/sessions`, reusing the existing `pi_session_jsonl` format — no new provider or source format needed since the on-disk schema is identical to `~/.pi/agent/sessions`.

This mirrors the `OPENCLAW_DEFAULTS` precedent, which already lists three default locations (`.openclaw`, `.clawdbot`, `.moltbot`) under a single provider. `discover_provider_sources_for_spec` and `default_location_import_probe` key off the provider (not the individual location), so both Pi default locations are probed/discovered/deduped automatically with no other code changes.

Follows up #40/#41 (0.19 Pi tree discovery support). No overlap with draft #43 (env-var-based custom directory discovery) — verified against its diff; it adds a new match arm elsewhere in `discover_provider_sources_for_spec` and never touches the `PI_DEFAULTS` constant.

## Validation

- Real-world OMP install: 326 sessions / 40k+ events previously importable only via explicit `--path`; discovery now finds the source natively.
- Added a regression test in `provider_sources.rs` mirroring the existing `.pi/agent/sessions` assertions (empty → available transitions, correct `source_format`) for the new `.omp/agent/sessions` location.
- Updated `docs/providers.md`, `docs/provider-support.md`, and `docs/provider-support-matrix.json` per the "Required Evidence For Promotion" checklist.
- `cargo test -p ctx-history-capture`: 86 passed, 1 ignored.
- `cargo test -p ctx-history-core provider_support`: 2 passed (matrix JSON still parses/matches schema).
- `cargo clippy -p ctx-history-capture --all-targets -- -D warnings`: clean.
- `cargo fmt`: clean.

### DIFF
--- a/crates/ctx-history-capture/src/provider_sources.rs
+++ b/crates/ctx-history-capture/src/provider_sources.rs
@@ -104,11 +104,18 @@ const CODEX_DEFAULTS: &[ProviderDefaultLocation] = &[
     },
 ];
 
-const PI_DEFAULTS: &[ProviderDefaultLocation] = &[ProviderDefaultLocation {
-    path_components: &[".pi", "agent", "sessions"],
-    source_format: "pi_session_jsonl",
-    source_kind: ProviderSourceKind::NativeHistory,
-}];
+const PI_DEFAULTS: &[ProviderDefaultLocation] = &[
+    ProviderDefaultLocation {
+        path_components: &[".pi", "agent", "sessions"],
+        source_format: "pi_session_jsonl",
+        source_kind: ProviderSourceKind::NativeHistory,
+    },
+    ProviderDefaultLocation {
+        path_components: &[".omp", "agent", "sessions"],
+        source_format: "pi_session_jsonl",
+        source_kind: ProviderSourceKind::NativeHistory,
+    },
+];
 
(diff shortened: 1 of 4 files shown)


msokh:
@luca-ctx mind taking a look? Follows the same pattern as #41.

luca-ctx:
Sorry forgot to comment back here to say thanks for the contribution @msokh!

PR #53, opened 2026-07-05
Clarify installer PATH setup

### LINKED ISSUE #51: Clarify PATH setup after Unix installer places ctx in ~/.local/bin
After running the Unix installer, `ctx` may not be available as a bare command if `~/.local/bin` is not on the user's `PATH`.

Observed locally:

```sh
curl -fsSL https://ctx.rs/install | sh
command -v ctx
# no result

~/.local/bin/ctx --version
# ctx 0.19.0
```

The installer successfully installs the binary to `~/.local/bin/ctx`, but users can reasonably interpret `ctx: command not found` as an install failure.

Suggested improvement:

- Call out the default install directory in README/getting-started docs.
- Ensure the installer prints a clear note when the install directory is not on `PATH`.

I searched existing issues and PRs for `PATH`, `.local/bin`, and installer-related terms and did not find an existing report for this specific problem.

kj-9:
## Summary

- Document that the Unix installer places `ctx` in `~/.local/bin` by default.
- Tell users to run `~/.local/bin/ctx` directly or add `~/.local/bin` to `PATH` when `ctx` is not found.
- Add a PATH warning to `scripts/install.sh` after installing the binary.

## Why

A successful install can look like a failure when the user's shell cannot resolve `ctx` because `~/.local/bin` is not on `PATH`.

Closes #51.

## Validation

- `bash -n scripts/install.sh`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ ctx indexes those logs into SQLite on your machine, then gives current and futur
 curl -fsSL https://ctx.rs/install | sh
 ```
 
+The Unix installer places the binary in `~/.local/bin` by default. If your
+shell cannot find `ctx` after installation, run `~/.local/bin/ctx` directly or
+add `~/.local/bin` to your `PATH`.
+
 Optional but recommended for agent sessions:
 
 ```bash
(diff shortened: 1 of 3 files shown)


luca-ctx:
Superseded by https://github.com/ctxrs/ctx/pull/58 which adds automatic PATH handling and info in the troubleshooting doc.

PR #54, opened 2026-07-05
fix(cli): create parent directories for docs show output

atharva-patill:
## Summary

`ctx docs show --out` currently fails when the output path contains directories that don't already exist.

This mirrors the existing behavior of `ctx show ... --out` and `ctx docs man --out` by creating parent directories before writing the output file.

## Reproduction

```bash
tmp=$(mktemp -d)
cd "$tmp"

/path/to/ctx/target/debug/ctx docs show \
  cli-reference \
  --out nested/doc.txt
```

Before:

Screenshot 2026-07-05 at 1 18 43 PM

After:

Screenshot 2026-07-05 at 1 30 17 PM

## Validation

- Added a regression test covering nested output paths.
- Verified the behavior manually.

### DIFF
--- a/crates/ctx-cli/src/docs.rs
+++ b/crates/ctx-cli/src/docs.rs
@@ -422,6 +422,9 @@ fn show_doc(args: DocsShowArgs) -> Result<()> {
         }
     };
     if let Some(path) = args.out {
+        if let Some(parent) = path.parent().filter(|parent| !parent.as_os_str().is_empty()) {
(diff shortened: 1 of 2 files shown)


luca-ctx:
Good catch, thank you! Just added fmt

atharva-patill:
> Good catch, thank you! Just added fmt

Thanks! That’s on me—I missed the formatting step.

PR #55, opened 2026-07-05
fix(cli): normalize whitespace-only search filters

atharva-patill:
## Summary

While looking at the search filters, I noticed that whitespace-only values passed to `--file` and `--workspace` were being ignored by the search pipeline, but they still appeared in the JSON `filters` output.

This normalizes both fields when building `SearchFilters`, so the JSON output reflects the filters that are actually applied.

## Reproduction

```bash
./target/debug/ctx search foo --file " " --json | jq '.filters'
```

Before:

```json
{
  "file": " ",
  "include_subagents": false
}
```

After:

Whitespace-only values are normalized away, so the `file` filter is no longer included in the JSON output. The same behavior now applies to `--workspace`.
Screenshot 2026-07-05 at 6 42 57 PM

## Testing

- Added a regression test covering both `--file` and `--workspace`.
- Verified the updated JSON output manually.

### DIFF
--- a/crates/ctx-cli/src/main.rs
+++ b/crates/ctx-cli/src/main.rs
@@ -6423,7 +6423,9 @@ fn search_filters(
         provider_key: source_identity.provider_key,
         source_id: source_identity.source_id,
         source_format: source_identity.source_format,
-        repo: input.workspace,
+        repo: input
+            .workspace
+            .and_then(|s| if s.trim().is_empty() { None } else { Some(s) }),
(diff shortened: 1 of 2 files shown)


luca-ctx:
Added fmt, thanks for this!

atharva-patill:
> Added fmt, thanks for this!

Thanks! I missed the cargo fmt. I’ll make sure to run it first next time.

PR #62, opened 2026-07-06
Split oversized modules and add LOC gate

luca-ctx:
## Summary
- Split oversized source and test modules into domain-focused Rust/Swift modules while preserving the newly added provider coverage from origin/main.
- Added a 650-line source/test LOC gate and wired it into Bazel fast checks.
- Updated provider support matrix scanning for path-based Rust modules used by the split.

## Verification
- bash scripts/check-loc.sh
- python3 scripts/check-provider-support-matrix.py
- cargo check --workspace --locked
- cargo clippy --workspace --all-targets --locked -- -D warnings
- bash scripts/check.sh --mode=fast
- bash scripts/bazel-test.sh cargo_test_default


luca-ctx:
Closing this draft for now. The current split is too mechanical/metric-driven and does not meet the architectural-cleanup bar; we are going to rebuild from origin/main with deliberate domain boundaries and a less perverse LOC policy.

PR #63, opened 2026-07-06
fix(opencode): import actual session messages, not just metadata

robbe1912:
## Summary

- \opencode_session_messages()\ short-circuits on \session_message\ having any rows, but in current OpenCode schema this table only stores agent-switched/model-switched metadata
- Actual conversation lives in \message\ table with \data.role\ = user/assistant — never reached
- Fix: guard the \session_message\ path by checking for known conversation entry types before accepting rows; fall through to \message\ table when only metadata exists

## Validation

- \cargo build --release\ (Rust 1.96.1, Windows)
- Re-imported: \ctx import --provider opencode --reset-cursor\
- Before: 112 notice events, 65 sessions
- After: 53678 message events (45521 assistant + 8157 user), 1511 sessions
- \ctx search\ returns real message content


luca-ctx:
Thanks for this fix!

PR #66, opened 2026-07-06
[codex] Render readable MCP tool content

luca-ctx:
## Summary

Render successful ctx MCP tool results as concise, semantic text in `content[0].text` while keeping `structuredContent` exact/lossless JSON.

This replaces the old generic structured-JSON stub with readable blocks for `status`, `sources`, `search`, `sql`, `show_session`, and `show_event`. The text output uses compact key-value summaries, capped Markdown tables, numbered search results, next-command hints, and short transcript snippets so MCP clients that only consume text still get useful context without dumping raw JSON.

## Validation

- `cargo fmt -- --check`
- `git diff --check`
- `cargo test -p ctx mcp_ --test cli`
- Manual MCP roundtrip against a temp fixture-backed store for `status`, `sources`, `sql`, `search`, `show_session`, and `show_event`; inspected `content[0].text` examples for readable key-value/table/result/transcript output.

## Notes

`structuredContent` remains the complete structured result. Text rendering is intentionally capped for token use, so agents should use `structuredContent` when they need every row, result, source, or transcript event.

### DIFF
--- a/crates/ctx-cli/src/mcp.rs
+++ b/crates/ctx-cli/src/mcp.rs
@@ -28,6 +28,14 @@ use super::{
 const MCP_PROTOCOL_VERSION: &str = "2025-11-25";
 const MCP_MAX_LINE_BYTES: usize = 1024 * 1024;
 const MCP_MAX_SESSION_EVENTS: usize = 200;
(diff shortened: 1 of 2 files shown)


luca-ctx:
Superseded by #64, which now carries the readable MCP content commit on top of the contributor's original fix.

PR #70, opened 2026-07-06
feat(cli): add --quiet flag to suppress non-essential output

robbe1912:
## Summary

- Add `--quiet` global flag to `Cli` that suppresses setup get-started tips, import details, and status info
- Human output for `ctx setup` and `ctx status` is skipped when `--quiet` is passed; JSON output and progress lines are unaffected
- `--quiet` is a global flag available on any subcommand and also set via `CTX_QUIET` env var

## Validation

- `cargo check -p ctx-history-core`
- Manual: `ctx --quiet setup` suppresses get-started block but still prints progress/status
- Manual: `ctx setup --quiet` (flag after subcommand) works via clap global flag propagation

### DIFF
--- a/crates/ctx-cli/src/commands/setup.rs
+++ b/crates/ctx-cli/src/commands/setup.rs
@@ -26,6 +26,7 @@ pub(crate) fn run_setup(
     args: SetupArgs,
     data_root: PathBuf,
     analytics_properties: &mut AnalyticsProperties,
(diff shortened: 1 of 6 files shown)


luca-ctx:
Thanks for the contribution! I did some cleanup on the quiet-mode contract, added coverage/docs, rebased it onto current main, and merged it.

PR #71, opened 2026-07-06
refactor(upgrade): avoid redundant Vec clone with take

robbe1912:
## Summary

- Replace 4 `plan.warnings.clone()` calls with `std::mem::take(&mut plan.warnings)` in `check_upgrade` and `apply_upgrade`
- The `plan` struct is consumed by `UpgradeOutcome` immediately after each clone, so taking the `Vec<String>` out avoids an allocation while preserving semantics
- `plan.warnings` becomes an empty `Vec` in the stored plan, which is fine since warnings are also present in the outcome's top-level `warnings` field

## Validation

### DIFF
--- a/crates/ctx-cli/src/upgrade/command.rs
+++ b/crates/ctx-cli/src/upgrade/command.rs
@@ -206,7 +206,7 @@ fn check_upgrade(
     channel_override: Option<&str>,
     command: &'static str,
 ) -> Result<UpgradeOutcome> {
-    let plan = build_upgrade_plan(config, channel_override, false)?;
+    let mut plan = build_upgrade_plan(config, channel_override, false)?;
     write_state_checked(data_root, &plan, "checked")?;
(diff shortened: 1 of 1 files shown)


luca-ctx:
Thanks for the cleanup. I’m going to close this one out.

The warnings in this path are meaningful user-facing upgrade diagnostics: they tell users about things like PATH shadowing, multiple ctx binaries, or an install marker problem where an upgrade may appear to work but the shell might still run a different binary afterward.

Given that, I’d rather keep the current code boring and explicit. The existing clone preserves the invariant that the plan still contains the warnings that the outcome reports. `std::mem::take` saves a tiny allocation on a non-hot path, but leaves `plan.warnings` empty while the outcome still has top-level warnings, which makes the internal state less obvious for very little benefit.

So the visible behavior is probably unchanged today, but the tradeoff is not worth the churn here.

PR #72, opened 2026-07-06
fix(tests): use platform-agnostic temp dirs instead of hardcoded Unix paths

robbe1912:
## Summary

- Replace hardcoded `/tmp/`, `/home/tester/`, `/repo/` paths with `tempfile::tempdir()` / `std::env::temp_dir()` in test files so tests pass on Windows
- Fixed files:
  - `crates/ctx-cli/src/install_marker.rs` — `appends_marker_suffix_to_full_exe_path` uses tempdir
  - `crates/ctx-cli/src/skill/tests.rs` — three path-resolution tests use tempdir-based home/repo dirs and `ends_with(Path::new(…))` assertions for cross-platform compatibility
  - `crates/ctx-history-core/src/tests.rs` — `local_layout_paths_are_flat_under_data_root` and `ctx_data_root_env_is_the_ctx_root_itself` use `std::env::temp_dir()`

## Validation

- `cargo check -p ctx-history-core`
- `cargo test -p ctx --bin ctx -- "skill::tests::"` — 9 pass, 0 fail
- `cargo test -p ctx --bin ctx -- "install_marker::"` — 4 pass, 0 fail
- `cargo test -p ctx-history-core -- "local_layout"` — 1 pass, 0 fail
- `cargo test -p ctx-history-core -- "ctx_data_root"` — 1 pass, 0 fail

### DIFF
--- a/crates/ctx-cli/src/install_marker.rs
+++ b/crates/ctx-cli/src/install_marker.rs
@@ -88,9 +88,11 @@ mod tests {
 
     #[test]
     fn appends_marker_suffix_to_full_exe_path() {
+        let temp = tempfile::tempdir().unwrap();
+        let exe = temp.path().join("ctx.exe");
         assert_eq!(
-            install_marker_path(Path::new("/tmp/ctx.exe")),
-            PathBuf::from("/tmp/ctx.exe.install.json")
+            install_marker_path(&exe),
+            temp.path().join("ctx.exe.install.json")
         );
     }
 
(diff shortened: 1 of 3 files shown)


luca-ctx:
Thanks for taking a pass at making these tests more platform-friendly. I’m going to close this one for now.

I checked the specific tests touched here against the Windows target, and the current `origin/main` versions already pass. Rust `Path`/`PathBuf` accepts `/` separators on Windows in these cases, so the hardcoded-looking Unix paths are not currently causing the demonstrated failures.

There are also a couple of tradeoffs in this patch that I’d rather avoid: it introduces unused-import warnings in test builds, and one assertion changes from checking the exact repo-relative target path to only checking `ends_with(...)`, which makes the test a bit weaker.

I’d be happy to reconsider a smaller cleanup if there is a concrete failing Windows case or if the assertions stay exact and the test build stays warning-free. For now, though, this does not look like a real issue we need to fix.

PR #80, opened 2026-07-06
Add linux-aarch64 (ARM64) support

shinhyo:
## What

Add linux-aarch64 support to `install.sh`, the public-CLI build script, and the Buildkite pipeline.

## Why

Currently `ctx` only ships linux-x64 binaries. On ARM64 Linux hosts (AWS Graviton, Apple Silicon Linux VMs, Raspberry Pi, etc.) the installer fails with:

```
error: cannot detect this host platform; set CTX_PLATFORM
curl: (23) Failure writing output to destination
```

with no release artifact to download.

## Changes

1. **`scripts/install.sh`** — detect `Linux:aarch64` and `Linux:arm64`, allow `linux-aarch64` in the platform allow-list, and document it in `--platform` help text.
2. **`scripts/build-public-cli-artifact.sh`** — add `linux-aarch64` case (target `aarch64-unknown-linux-gnu`, artifact name `ctx-linux-aarch64`) and reuse the existing `cargo build` path (no zig/cross needed on a native aarch64 host).
3. **`.buildkite/pipeline.yml`** — add a public-CLI build step gated on `CTX_PUBLIC_CLI_ARTIFACT_MATRIX`, targeting the existing `release-linux-managed` queue with `arch: arm64`.

## Testing

- Built locally on an aarch64 Linux host: `scripts/build-public-cli-artifact.sh linux-aarch64` produces a working `ctx-linux-aarch64` binary.
- That binary runs `ctx setup` successfully, indexing Pi / Codex / OpenCode / Hermes / Gemini sessions (19 sessions, 1,408 events).
- `ctx search` and `ctx show event` confirmed against the same data.
- `install.sh` now auto-detects `aarch64` instead of erroring.

## Notes for maintainers

- The pipeline step is gated by `CTX_PUBLIC_CLI_ARTIFACT_MATRIX`, same as the other public-CLI targets.
- `arch: arm64` will require either a native arm64 Buildkite agent in the `release-linux-managed` queue, or cross-tooling on existing x86_64 agents (gcc-aarch64-linux-gnu + libc6-dev-arm64-cross). Happy to wire up zig-based cross-compilation like the macOS targets if that is preferred.

---

*Tested and pushed from an aarch64 Linux host.*

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -77,6 +77,22 @@ steps:
       - "target/public-cli-artifacts/ctx.sha256"
       - "target/public-cli-artifacts/ctx.version"
 
+  - label: ":package: public CLI linux-aarch64"
+    key: "public-cli-linux-aarch64"
+    if: build.env("CTX_PUBLIC_CLI_ARTIFACT_MATRIX") == "1"
+    command: |
+      scripts/build-public-cli-artifact.sh linux-aarch64
+    agents:
+      queue: "release-linux-managed"
+      ctx-runner-class: "release-linux-control"
+      os: "linux"
+      arch: "arm64"
+    timeout_in_minutes: 60
+    artifact_paths:
+      - "target/public-cli-artifacts/ctx-linux-aarch64"
+      - "target/public-cli-artifacts/ctx-linux-aarch64.sha256"
+      - "target/public-cli-artifacts/ctx-linux-aarch64.version"
+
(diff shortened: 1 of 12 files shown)


luca-ctx:
Thanks for this! Added some more robustness in other locations in the release system for arm64 and merged.

0.21, coming within the next 24 hours more than likely, will be with arm64 support.

luca-ctx:
0.21 is released with arm support, let me know how it goes!

PR #89, opened 2026-07-07
feat(capture): emit synthetic events for oversized SQLite rows

robbe1912:
## Summary

Builds on #88. Phase 1 made oversized SQLite rows **not abort** the source import by skipping them at the SQL layer. The skipped rows themselves were invisible — no event landed in the store, no search hit referenced them, `ctx show session` silently omitted them. For providers like the Kilo Code VSCode extension that occasionally carry a single multi-hundred-MB message (e.g. a giant multi-file diff captured by the extension), that meant an entire legitimate session could be silently dropped from history.

This change re-imports each oversized row as a **synthetic event** carrying a truncated preview of the original payload plus a `_ctx_oversized` marker that records the original byte count and source table. The session stays searchable end-to-end, and downstream consumers can detect truncation explicitly.

- Add `truncated_oversized_message_rows` that opens an unbounded read-only connection (the bounded connection refuses to materialize the value, so truncation must run against an uncapped connection — same reasoning as `oversized_message_row_ids` in #88) and emits one `OpenCodeMessageRow` per oversized row, with `data` set to a synthetic JSON payload containing the truncated preview (`PROVIDER_MAX_TEXT_CHARS` chars) and a `_ctx_oversized` marker.
- Extend all three row fetchers (`opencode_session_message_rows`, `opencode_session_entry_rows`, `opencode_message_rows`) to append the synthetic rows before returning so they flow through the existing per-row import path unchanged.
- Drop the Phase 1 `summary.skipped` increment in `normalize_opencode_sqlite`: oversized rows are no longer skipped, they are imported as synthetic events and counted as `imported_events`.
- Surface the `_ctx_oversized` marker in `opencode_event.payload` as a top-level field alongside `body` for downstream visibility. The marker is also nested inside `body` but `provider_capped_json` would otherwise lose it for any oversized row whose synthetic preview also exceeds the body cap, so the top-level field is the canonical detection point.

Marker shape (at `event.payload.body.ctx_oversized`):

```json
{
  "original_bytes": <byte count of the oversized row>,
  "preview_chars": 16000,
  "source_table": "session_message",
  "source_format": "opencode_sqlite",
  "preview_truncated": false
}
```

Affected scope: only the OpenCode-family SQLite adapter (covers the Kilo Code extension via `KiloSqliteAdapter`). JSONL adapters are unaffected.

## Validation

- `cargo check -p ctx-history-capture` — clean
- `cargo build --workspace` — clean, only the 2 pre-existing warnings on the `ctx` bin (`ApplyResult::Applied` unused, `ProcessState::Running/NotRunning` never constructed) that exist on `main`
- `cargo test -p ctx-history-capture --lib tests::native_sqlite` — 29 passed, 0 failed
- Regression test `native_opencode_truncates_oversized_sqlite_text_value_and_preserves_other_rows` asserts:
  - `summary.failed == 0` (oversized rows must not be treated as failures)
  - `summary.imported_events >= 1` (other rows still import)
  - exactly one event in the session carries a non-null `payload.body.ctx_oversized`
  - `payload.body.ctx_oversized.original_bytes` matches the oversized row size
  - `payload.body.ctx_oversized.source_table` / `source_format` are set
  - `payload.body.message_id` preserves the original provider row id

## End-to-end verification

Built ctx from this branch and imported a synthetic `opencode.db` fixture (smoke-test schema) containing 2 normal rows + 1 oversized row (`MAX_PROVIDER_SQLITE_VALUE_BYTES + 1` bytes under `text`):

```
[ctx] imported: events=3, sessions=1, failed=0, skipped=0
[store] exactly one synthetic oversized event with _ctx_oversized marker
[store]   message_id preserved
[store]   original_bytes matches oversized row size
[store]   source_table=session_message, source_format=opencode_sqlite
[PASS] PR #88 + #89 verified end-to-end through ctx CLI
```

All three rows imported cleanly (Phase 1 effect: no `SQLITE_TOOBIG` abort), and exactly one synthetic event landed in the store with the `_ctx_oversized` marker at `payload_json.body.ctx_oversized` (Phase 2 effect).

## Depends on

- #88 (Phase 1: skip oversized SQLite rows instead of aborting source). This branch is based on top of `fix/search-refresh-oversized-skip` and should land after it.

## Related

- #90 tracks the separate Kilo Code dialect gap (schema lacks `text`/`content`/`message` keys, blocks all kilo imports regardless of oversized handling).

### DIFF
--- a/crates/ctx-history-capture/src/provider/providers/opencode.rs
+++ b/crates/ctx-history-capture/src/provider/providers/opencode.rs
@@ -10,7 +10,7 @@ use ctx_history_core::{
     ProviderRedactionBoundary, ProviderSessionEnvelope, ProviderSourceEnvelope,
     ProviderSourceTrust, RedactionState, SessionStatus, PROVIDER_CAPTURE_ENVELOPE_SCHEMA_VERSION,
 };
-use rusqlite::Connection;
+use rusqlite::{Connection, OpenFlags};
 use serde_json::{json, Value};
 
 use crate::compute_payload_hash;
@@ -27,8 +27,8 @@ use crate::provider::native::{
 use crate::provider::providers::real_content::text_has_real_content;
 use crate::{
     CaptureError, ProviderAdapterContext, ProviderNormalizationResult, Result,
-    KILO_SQLITE_SOURCE_FORMAT, OPENCODE_SQLITE_SOURCE_FORMAT, PROVIDER_MAX_PREVIEW_CHARS,
-    PROVIDER_MAX_TEXT_CHARS,
+    KILO_SQLITE_SOURCE_FORMAT, MAX_PROVIDER_SQLITE_VALUE_BYTES, OPENCODE_SQLITE_SOURCE_FORMAT,
+    PROVIDER_MAX_PREVIEW_CHARS, PROVIDER_MAX_TEXT_CHARS,
 };
 
 pub(crate) struct OpenCodeSqliteDialect {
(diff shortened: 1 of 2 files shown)


luca-ctx:
Thanks for the PR! I am going to resolve the oversized row issue on top of your other PR (#88) and close this one as we are going to do a different approach than synthetic events because those could give the impression that the synthetic data is real data. We would like to either mark the row as skipped for oversize, or somehow fix the oversize issue. Hopefully this makes sense and stay tuned on #88

PR #100, opened 2026-07-08
Make usage analytics opt-in with install-time disclosure

mtgibbs:
## Make usage analytics opt-in and disclose data collection up front

ctx indexes users' full agent conversation history — some of the most sensitive data on a developer's machine. The engineering already respects that (local-only indexing, no transcript egress, honest `docs/storage.md`), but the consent model doesn't match: analytics are default-on, the only disclosure lives in `docs/storage.md`, and `scripts/check-docs.sh` actively fails CI if the word "analytics" appears in the README. A user who runs `curl -fsSL https://ctx.rs/install | sh` starts transmitting install/device identifiers before they've had any chance to learn the behavior exists.

This PR aligns the consent model with the engineering:

### Changes

- **Analytics are now opt-in (default off).** `AppConfig::default()` sets `analytics.enabled = false`. A fresh install sends no events and creates no device identity.
- **New `ctx analytics enable | disable | status` subcommand**, mirroring `ctx upgrade enable/disable`. `enable` writes `[analytics] enabled = true` to `config.toml`; the command itself never emits an analytics event.
- **Installers disclose before acting and make opting in easy.** `scripts/install.sh` and `scripts/install.ps1` print a three-line `Privacy:` block (local index, opt-in analytics, daily auto-upgrade) before doing anything. Interactive `install.sh` runs ask once — `Share anonymous usage analytics to help improve ctx? [y/N]` via `/dev/tty`, defaulting to No; non-interactive runs stay off unless `--enable-analytics` / `CTX_INSTALL_ENABLE_ANALYTICS=1` (PowerShell: `-EnableAnalytics`).
- **README gains a "Private by default" section** stating the contract in plain language, with a pointer to `docs/storage.md` for the full payload details.
- **`docs/storage.md` and `docs/cli-reference.md` updated** from "default-on, here's the opt-out" to "off by default, here's how to opt in."
- **`scripts/check-docs.sh`** now permits analytics copy in `README.md` and `docs/cli-reference.md` (still banned everywhere else).
- **Tests:** all analytics integration tests now opt in explicitly (`CTX_ANALYTICS_ENABLED=true`); two new regression tests pin the contract — a default invocation sends nothing and creates no device identity, and the settings command round-trips `config.toml`.

### Not covered here (needs hosted-side changes)

The production hosted installer at `https://ctx.rs/install` is not in this repository. It currently POSTs stage-by-stage install diagnostics tagged with a per-download `install_attempt_id` before the user has seen any disclosure. Suggest porting the same pattern there: print the `Privacy:` block first, make the stage diagnostics opt-in (or at minimum gate them on the same prompt), and forward the analytics answer to the CLI.

### Why default-off matters for this tool in particular

For a general dev tool, default-on coarse analytics is a defensible norm. ctx is different: its entire value proposition is "point me at every secret-laden transcript on your machine, trust me that nothing leaves." The docs make exactly that promise ("local and private by default"). Default-on identifier transmission — however coarse — undercuts the trust the product depends on, and the users most likely to adopt ctx are precisely the ones most likely to notice. Opt-in costs some analytics volume, but an explicit install-time prompt should recover much of it while making every remaining data point consensual.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,16 @@ ctx is written in Rust and stores a local SQLite index, so searches are fast, sc
 
 The index is local and private by default. Transcript text is preserved rather than hiding local paths or secret-shaped strings, so review copied output before sharing it outside the machine.
 
+## Private by default
+
+Your agent history is sensitive, and ctx treats it that way:
+
+- Indexing, search, and SQL run entirely on your machine. Transcript content, search queries, and file paths never leave it.
+- Anonymous usage analytics are **opt-in and off by default**. If you want to help improve ctx, run `ctx analytics enable` — this shares coarse product metadata only (command name, success state, bucketed counts), never session text, queries, or paths. Opt back out anytime with `ctx analytics disable`.
+- Installer-managed binaries check for signed upgrades once a day. Disable with `ctx upgrade disable`.
+
+The full storage and privacy contract, including the exact analytics payload when enabled, is documented in [docs/storage.md](docs/storage.md).
+
 ```bash
 # Index all of your existing local agent sessions
 ctx setup
--- a/crates/ctx-cli/src/config.rs
+++ b/crates/ctx-cli/src/config.rs
@@ -34,7 +34,9 @@ impl Default for AppConfig {
     fn default() -> Self {
         Self {
             analytics: AnalyticsConfig {
-                enabled: true,
+                // Privacy default: analytics are opt-in. Enable with
+                // `ctx analytics enable` or `[analytics] enabled = true`.
+                enabled: false,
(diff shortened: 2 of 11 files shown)


luca-ctx:
Thanks for your suggestion and contribution, however we are going to continue with our current policy.

We are committed to privacy for transcripts, source code, etc and only collect opt-out analytics on coarse, pseudonymous usage metrics. If you want to opt out of these, the instructions are clear on how to do that in ~/.ctx/config.toml.

Of course, you are also welcome to fork and strip out any analytics plumbing since it is apache licensed.

PR #103, opened 2026-07-09
Preserve itemType in search citations

atharva-patill:
## Summary

`itemType` is part of the public `Citation` contract but was being omitted during Python SDK normalization.

This preserves the field in search citations.

## Testing

- Added a regression test covering search citation normalization.
- Verified the Python SDK test suite passes.


luca-ctx:
Thanks for catching this. I changed the fix to address the underlying overloaded itemType field and JSON shape instead of just preserving `itemType` in the Python SDK.

This now uses separate fields for those jobs: `payload_type`, `result_type`, `target_type`, and `record_type`. And it is fixed across all of the SDKs.

atharva-patill:
> Thanks for catching this. I changed the fix to address the underlying overloaded itemType field and JSON shape instead of just preserving `itemType` in the Python SDK.
> 
> This now uses separate fields for those jobs: `payload_type`, `result_type`, `target_type`, and `record_type`. And it is fixed across all of the SDKs.

Thanks for looking into it and for the broader fix!

PR #113, opened 2026-07-09
Harden semantic search platform support

luca-ctx:
## Summary

- expands local semantic backend support to Linux GNU x64 and aarch64 while keeping macOS, Windows GNU, and FreeBSD lexical-safe
- routes semantic query embedding through a daemon-owned authenticated query endpoint instead of foreground model initialization
- preserves daemon-owned lexical refresh on unsupported semantic platforms and makes hybrid/default search fall back with `unsupported_platform`
- hardens daemon status/lock handling, including Windows process liveness checks and query endpoint cleanup
- updates docs for daemon query-service behavior, platform limits, and setup/search read-write semantics

## Validation

- `cargo fmt --check`
- `cargo check --workspace --locked`
- `cargo check -p ctx --target x86_64-pc-windows-gnu --locked`
- `cargo test -p ctx query_service_transport_tests --locked`
- `cargo test -p ctx hybrid_search_reports_missing_daemon_query_service --locked`
- `cargo test -p ctx --test index index_status_discovers_existing_semantic_model_cache_dirs --locked`
- `cargo test -p ctx daemon_report_marks_orphaned_running_status_recoverable --locked`
- `cargo test -p ctx --test setup_sources_import setup_backgrounds_discovered_codex_sessions_when_daemon_is_enabled_and_wait_imports --locked`
- `scripts/check-docs.sh`
- `scripts/check-buildkite-pipeline.sh`
- `git diff --check`
- isolated `target/debug/ctx` setup/daemon/search smoke
- `CTX_PUBLIC_CLI_ARTIFACT_DIR=target/public-cli-artifacts-smoke scripts/build-public-cli-artifact.sh linux-x64`
- isolated staged artifact setup/daemon smoke

## Notes

Local Bazel public-smoke wrapper was attempted for `//:cargo_check //:cargo_fmt_check //:docs_check`, but the local run failed before target analysis because `rules_rust` rejects the parent Cargo config at `/home/daddy/.cargo/config.toml`. The underlying cargo/docs/pipeline checks listed above passed directly.

This PR does not claim true semantic support on macOS/Windows/FreeBSD yet. Those artifacts remain daemon/lexical-safe until the ONNX/vector backend and query transport are proven for each target.


luca-ctx:
Superseded by #132, which rebases the complete opt-in semantic daemon work onto current main, removes the public real-corpus validation summary, and includes the final cross-platform runtime/release hardening.

PR #118, opened 2026-07-10
Normalize lookup identifiers in TypeScript SDK

atharva-patill:
## Summary

Normalize lookup identifiers before invoking the local CLI.

This aligns the TypeScript SDK with the CLI by:

- trimming lookup identifiers before transport
- rejecting whitespace-only identifiers
- forwarding the normalized value

Adds regression tests covering both whitespace-only and whitespace-padded inputs.

### DIFF
--- a/sdks/typescript/src/index.js
+++ b/sdks/typescript/src/index.js
@@ -174,8 +174,8 @@ export class LocalAgentHistoryClient {
   }
 
   async showEvent(id, options = {}) {
-    requireId("event id", id);
-    const args = ["show", "event", id, "--format", "json"];
+    const normalizedId = normalizeLookupId("event id", id);
+    const args = ["show", "event", normalizedId, "--format", "json"];
     appendOptionalNumber(args, "--before", options.before);
     appendOptionalNumber(args, "--after", options.after);
     appendOptionalNumber(args, "--window", options.window);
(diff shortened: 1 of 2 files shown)


luca-ctx:
Thanks for the contribution. Lookup identifier normalization is intentionally owned by the ctx runtime so behavior remains consistent across all SDKs and CLI versions. Since padded identifiers are already normalized and whitespace-only identifiers are rejected by the runtime, we don’t want to duplicate that logic in individual SDKs. Closing for now.

Although let me know if you have experienced specific issues with this that make it a problem worth solving.

atharva-patill:
> Thanks for the contribution. Lookup identifier normalization is intentionally owned by the ctx runtime so behavior remains consistent across all SDKs and CLI versions. Since padded identifiers are already normalized and whitespace-only identifiers are rejected by the runtime, we don’t want to duplicate that logic in individual SDKs. Closing for now.

Thanks for the clarification! I had interpreted it as a CLI/SDK contract mismatch, but understanding that normalization intentionally lives in the runtime changes the picture.

> Although let me know if you have experienced specific issues with this that make it a problem worth solving.

I haven’t run into a user-facing issue myself—I found it during a consistency audit. Thanks for explaining the design decision.

PR #139, opened 2026-07-12
Bound FTS merge recovery steps on large indexes

j2h4u:
## What changed

Reduce the explicit FTS5 merge page budget from 1,024 to 16 pages per committed/checkpointed step.

## Why

The FTS5 merge page parameter is a work budget, not a hard bound on physical WAL pages. On a real 2.3 GiB ctx database populated from a 206 MiB Hermes source, one 1,024-page recovery step grew the WAL to 7.0 GiB before the process was stopped for disk safety. This also made ordinary `ctx search` appear hung because writable store open synchronously resumes the recovery marker.

The smaller step keeps the existing crash-safe marker, serialization, merge semantics, and strict checkpoint loop unchanged while reducing the amount of FTS work that can occur inside one statement.

## Real-corpus validation

Reused the exact interrupted database and `event_search_bulk_mode_v1` marker that reproduced the failure:

- before: 1,024-page step, WAL reached 7.0 GiB and did not finish safely
- after: 16-page step, recovery completed in 1m52s
- measured peak WAL: 14,959,752 bytes (~14.3 MiB)
- final WAL: 0 bytes
- recovery marker: cleared
- `ctx search Hermes --limit 3`: returned results, including a Hermes event
- exit status: 0

## Checks

- `cargo test -p ctx-history-store connection_tests -- --nocapture` (7 passed)
- `cargo clippy -p ctx-history-store -p ctx --tests -- -D warnings`
- `cargo build -p ctx --release`

## Search latency follow-up

The same recovery marker also made ordinary search look hung because lexical event ranking re-joined the FTS5 virtual table after materializing its matches. This branch now keeps the match rows in a materialized CTE and ranks them without that second FTS scan. Scriptgram queries retain the existing path.

Live validation on the existing 2.3 GiB database after installing the release binary:

- `ctx search Hermes --refresh off`: 0.70–1.03s over repeated runs
- `ctx search Hermes --provider hermes --refresh off`: 0.69–0.72s
- WAL remained absent/0 bytes and no recovery marker remained

Additional checks: `cargo fmt --check`, `cargo test -p ctx-history-store search -- --nocapture` (38 passed), clippy, and release build.

## Idempotent refresh follow-up

Background search was also re-normalizing unchanged non-manifest native roots after inventory had already marked them indexed. This was especially expensive for live SQLite sources: Hermes (about 216 MiB), OpenCode (about 675 MiB), and OpenClaw (about 350 MiB). The import path now checks the persisted root inventory and skips provider normalization when the source metadata is unchanged and there is no pending import file. Full rescans and changed/pending sources still run normally.

Validation:

- regression test: `unchanged_root_source_skips_provider_normalization`
- unchanged Hermes source: plain `ctx search Hermes --provider hermes --limit 1` returned in 1.11s
- generic plain search completed in 20.95s while reporting six pre-existing malformed Codex records; it did not hang and final WAL was absent
- release binary installed locally from commit `31a4dfdf`

### DIFF
--- a/crates/ctx-cli/src/commands/import/native.rs
+++ b/crates/ctx-cli/src/commands/import/native.rs
@@ -72,6 +72,19 @@ pub(crate) fn import_one_source_inner(
     let record = import_record_for_source(source);
     let record_id = record.id;
     store.upsert_record(&record)?;
+    // Non-manifest SQLite/tree sources still have a persisted root inventory.
+    // Do not normalize an unchanged, already-indexed source just because a
+    // background search discovered it again; those adapters may read the
+    // entire provider database before deduplication can occur.
+    if !full_rescan
+        && !source_uses_import_file_manifest(source)
+        && preinventory.source_root_file().is_some()
+        && store
+            .list_pending_source_import_files(source.provider, &source.path.display().to_string())?
+            .is_empty()
+    {
+        return Ok(ProviderImportSummary::default());
+    }
     let summary = if !full_rescan && source_uses_import_file_manifest(source) {
         import_manifested_source(
             store,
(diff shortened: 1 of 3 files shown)


j2h4u:
Follow-up after installing commit `9561d35` locally: the WAL-budget fix is now accompanied by the independent lexical-search fix.

The old ranking query scanned `event_search` once to find matches and then re-scanned the FTS5 virtual table while joining those rows back for ranking. On the populated Hermes database that made `ctx search` look hung after recovery. The new path materializes the FTS matches once and ranks that relation; scriptgram queries keep the existing mixed path.

Measured on the live 2.3 GiB database: repeated `ctx search Hermes --refresh off` runs took 0.70–1.03s, and provider-filtered runs took 0.69–0.72s. WAL stayed absent/0 bytes and the recovery marker stayed cleared. The branch remains draft while this is reviewed.

j2h4u:
Third follow-up after commit `31a4dfdf`: ordinary background refresh was repeatedly normalizing unchanged non-manifest roots even when `source_import_files` already said they were indexed. That made a normal search re-read large live SQLite sources (Hermes ~216 MiB, OpenCode ~675 MiB, OpenClaw ~350 MiB). The import path now skips only unchanged, already-indexed roots; changed/pending sources and explicit/full rescans remain untouched.

On the live host, repeated plain Hermes search after the initial refresh returned in 1.11s with no WAL. A generic plain search completed in 20.95s and reported six existing malformed Codex records instead of hanging; WAL was absent afterward. The branch remains draft for review.

luca-ctx:
This mixed draft has now been fully integrated through focused, independently reviewed PRs while preserving @j2h4u as the commit author:

- #140 — bounded FTS merge recovery/finalization on populated indexes
- #142 — single-scan lexical event search with event-ID ranking and query-plan regressions
- #144 — unchanged native search-refresh fast path with source-format invalidation and WAL-aware freshness tokens

Splitting the concerns let each change keep a narrow contract and full regression coverage. The final native-refresh design is intentionally search-refresh-only so explicit import/full-rescan semantics remain unchanged, and it detects committed WAL-only changes even when the main SQLite file's size and mtime do not move.

Closing this draft as superseded by those merged PRs. Thank you for the real-corpus diagnosis, measurements, and original implementations.

PR #153, opened 2026-07-13
Bound FTS optimization merge steps

j2h4u:
## Draft: shared WAL-safety work in progress

This PR is deliberately still a draft. It generalizes FTS merge suppression across normalized provider imports and removes automatic full-index compaction after imports.

Live validation exposed a remaining common issue: normal imports still wrap an entire provider source in one SQLite transaction, so a large source can grow the WAL even with FTS merges suppressed. The next revision will make the existing bounded import transaction behavior the normal shared path (not a Hermes-/partial-only exception), while preserving crash-safe resume semantics.

Do not review or merge yet; I will update this description and tests after that final shared change is validated.

### DIFF
--- a/crates/ctx-cli/src/commands/import.rs
+++ b/crates/ctx-cli/src/commands/import.rs
@@ -673,11 +673,6 @@ pub(crate) fn run_import_internal(
         }
     }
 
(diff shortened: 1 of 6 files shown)


j2h4u:
Superseded by the narrower partial-only DRY fix in the new PR. Closing this draft so it cannot be reviewed or merged accidentally.

PR #154, opened 2026-07-13
Bound partial provider imports and avoid OpenClaw full FTS rebuild

j2h4u:
## Summary

- Use the existing bounded batch transaction path for all normalized provider imports when `--partial` is selected.
- Keep FTS merge bounds during those batches.
- Treat OpenClaw as incrementally searchable, so its import does not run an unnecessary full post-import FTS rebuild.
- Leave public flags and the default atomic import semantics unchanged.

## Validation

Live all-harness validation details are in the comment below.

### DIFF
--- a/crates/ctx-cli/src/commands/import/catalog.rs
+++ b/crates/ctx-cli/src/commands/import/catalog.rs
@@ -306,56 +306,10 @@ pub(crate) fn mark_catalog_sessions_failed(
 }
 
 pub(crate) fn source_uses_incremental_event_search(source: &SourceInfo) -> bool {
-    matches!(
-        source.provider,
-        CaptureProvider::Codex
-            | CaptureProvider::Claude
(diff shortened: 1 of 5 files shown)


j2h4u:
Final real-host validation is complete on a fresh database: `ctx import --all --partial --no-daemon` finished in 15m42s, with a 149.7 MiB peak WAL and final WAL of 0; searches succeeded for Codex, Claude, OpenCode, OpenClaw, and Hermes. The run separately reported 13 malformed Codex JSONL inputs, so this is not a claim that every source file is pristine.

That all-harness run also exposed an omitted classification: OpenClaw is incrementally searchable, so the final commit avoids its unnecessary full post-import FTS rebuild. Public flags and default atomic behavior remain unchanged.

We did not open a separate issue because the complete reproducer, fix, and validation evidence are contained in this PR.

PR #156, opened 2026-07-13
Preserve runtime-user provenance across deduplicated imports

j2h4u:
## Problem

ctx records the provider/runtime for imported history, but not the Unix user under which that runtime ran.

On a multi-user host, histories from the same provider in distinct user environments cannot be distinguished by downstream analysis. Deduplication can also observe one logical event from multiple user environments; retaining only one source loses that context.

## What this changes

Runtime remains the existing `provider`. This adds `ctx import --runtime-user` for the explicit Unix account under which the harness ran; it is neither the message author nor the user who runs `ctx import`.

Capture sources and sessions are scoped by that explicit user. Legacy rows remain `NULL`; migrations are additive and do not backfill.

## Design

An identical event observed for multiple runtime users remains one canonical event with per-source/session observations. For example, snapshots prepared for `alice` and `bob` can retain separate provenance without identity being inferred from paths.

Maintainer design question: is source-scoped `runtime_user` plus event observations the preferred shape, or would you prefer another integration point?

## Validation

Local verification: `cargo fmt --check`; workspace, capture, and store tests; the exact Codex fast-directory CLI test for `alice` + `bob`; and the Codex tail test.


luca-ctx:
Thanks so much for the contribution, however I will have to close this one. The design of the local CLI is deliberately single machine, single user. Once you start to introduce multiple users and multiple machines, the surface area expands significantly beyond what most people are looking for. However, we have a private beta for an optional cloud hosted version that works with teams (multi-user) and multi-machine sync. If you're interested in that you could check out https://ctx.rs/cloud/

Or if you really need the multiple local users thing as you have proposed it, I would encourage you to maintain a thin fork for yourself as I think it could meet your needs well.

Issue #60, opened 2026-07-06
Does past-session search improve agent security review? (early feedback)


xodn348:
You asked whether ctx / past-session search helps agents produce better security review work. Here is an early answer from actually testing it.

Yes, but at a specific layer. It does not improve raw detection (that is the scanner's job). What it improves is the agent's memory of prior findings and decisions:

- which issues were already found in this area,
- which were reviewed and accepted as deliberate exceptions (so the agent does not re-flag them or "fix" working code), and
- which fixes previously regressed.

Best used by querying history only when a scanner flags something, not by loading history at every agent start. That keeps token cost proportional to real findings instead of to history size. Concretely, this pairs well with a stateless scanner: run something like codesure on the changed code first, and only call ctx when the scan flags a problem, to pull the prior context for that specific finding.

One caveat specific to security: recalled transcript text re-enters a live agent's context, so it is a prompt-injection surface. In a security workflow especially, the agent should treat recalled history as data, not as instructions.

luca-ctx:
This is great feedback, thanks so much for taking the time to try it out and think deeply about it!

Firstly, I would agree with you that ctx / past session search does not imbue the agents with new capabilities, so we should not expect it to improve raw detection or raw performance on other tasks.

Secondly, that absolutely makes sense that you want to use targeted history searches only when you have something concrete and actionable to search. You don't really want to load all agent history for any task, even beyond security review.

And yes, anything an agent consumes is a prompt-injection surface. Interestingly, this is kind of like a "prompt injection replay" scenario where you could actually see the prompt injection a second time (the first time was when the agent first encountered it, and now you are searching back over it again).

I would be interested to know if any additional mitigations help, such as wrapping the results in markup tags like <search-123-these-are-not-instructions> {data} </search-123-these-are-not-instructions> where "123" is an unguessable value.

The supabase MCP server does a pretty elegant job of doing something like that, since you can potentially retrieve user-submitted data from it.

As of now, I would qualify new risk of prompt injection introduced to be low, since it is only introducing it as a replay of something it has already seen before. 

But if anyone is interested in piloting some experiments to test this, I would love to know the findings.

It would also be good to know the "sweet spot" for instructions for your agents as to when to use ctx / past session search as I can imagine you might want to tune it differently based on your project / task.

hampsterx:
[...]

PR #160, opened 2026-07-13
fix: nonce-tag recalled snippet delimiters to prevent boundary spoofing

kamleshchandela:
Related discussion: #60

## Summary

When the history-search feature recalls a past snippet and renders it back to an agent (CLI or MCP text output), the snippet is now wrapped in a delimiter carrying a random, single-use nonce. This makes it structurally impossible for the *content* of a recalled snippet to forge the delimiter that marks its own boundary.

## Why this matters

`ctx search` returns text that an agent reads and acts on. Some of that text — the recalled snippet itself — originates from data the system doesn't fully control (something said or stored earlier, possibly by another party). If there's nothing reliably marking "this part is recalled data, not a live instruction," a snippet could contain text that looks like the end of the recalled block, followed by what reads as a fresh instruction. An agent parsing the output has no reliable way to tell a real boundary from a fake one embedded in the data.

This is a boundary-spoofing / prompt-injection risk: the danger isn't that the recalled content is wrong, it's that the *structure* around it can be faked.

## How the fix works

Recalled snippets are now wrapped like this:

```
[[RECALLED_DATA nonce=3f9a2e7c-...]]
...snippet content, XML-escaped...
[[/RECALLED_DATA nonce=3f9a2e7c-...]]
```

- **Nonce is a random UUIDv4**, generated fresh per response. Because it's generated after the snippet content already exists in storage, nothing inside a snippet can pre-guess it and forge a matching delimiter.
- **One nonce per response, not per snippet.** All snippets in a single response share a nonce, so an agent can tell which ones came from the same recall call. This doesn't weaken the protection — an attacker can't predict a future random value whether it's scoped per-response or per-snippet — so sharing it is a readability win with no security cost.
- **The closing delimiter carries the nonce too, not just the opening one.** If only the opening tag were nonce-tagged, injected content could still forge a plain closing tag and escape the boundary early.
- **Content is clipped to the display length limit *before* wrapping, not after.** This guarantees the closing delimiter itself is never truncated — clipping after wrapping could cut a long snippet off mid-delimiter and leave the block unclosed in the rendered output.
- **Snippet content is XML-escaped** before wrapping, so special characters inside it can't be mistaken for markup.

## What this does *not* solve

This PR hardens the **boundary** only — it makes it structurally impossible for recalled content to impersonate a new instruction or forge a delimiter it didn't generate. It does **not** address the separate, deeper problem @hampsterx raised: an agent trusting a *claim* inside a recalled snippet (e.g. "this was already reviewed and approved") without checking that claim against something access-controlled. The nonce guarantees the box is real; it says nothing about whether the contents of the box are true. That's a distinct piece of work and a good candidate for a follow-up discussion.

## Scope / Compatibility

- Only touches **human/agent-facing text rendering** — CLI output and MCP text responses.
- The **`--json` path is untouched on purpose.** Scripted consumers (`ctx search --json | jq ...`) parse structured fields directly and don't face this ambiguity, so wrapping isn't needed there.
- **Note for reviewers:** this does change the visible shape of CLI/MCP text output (snippets are now wrapped). Anything downstream scraping the plain-text output with fixed assumptions — rather than using `--json` — should be checked against this change.
- `threat-model.md` is updated to document this mitigation and its boundaries.

## Testing

All of the following pass on this branch:

```bash
cargo build --workspace
cargo test -p ctx-history-search
cargo clippy -p ctx-history-search --all-targets -- -D warnings
cargo fmt --check
```

## Files changed (8)

| File | Role |
|---|---|
| `snippets.rs` | Core wrapping/escaping logic for recalled snippets |
| `lib.rs` | Wires the new wrapping logic into the public API |
| `tests/snippets.rs` | Unit tests for delimiter/nonce wrapping |
| `tests/mod.rs` | Test module registration |
| `mcp/text.rs` | MCP text-rendering integration |
| `search_render.rs` | CLI render integration |
| `commands/search.rs` | `ctx search` command wiring |
| `threat-model.md` | Documents the mitigation and its scope |

## Follow-up

Happy to open a separate discussion on the content-trust problem (verifying claims inside recalled data against an access-controlled source) if useful — keeping this PR scoped to the tagging mitigation that was already flagged as worth trying.

cc @hampsterx

---
Suggested labels: `security`, `enhancement`


luca-ctx:
Thanks so much for the contribution! 

However I don't think we're ready to merge this. I tried to reproduce a scenario locally where I could place a prompt injection attack in my own ctx history, and I couldn't actually trigger it.

We would need to prove that we can get a successful prompt injection attack in ctx history, and then mitigate it with the nonce.

Otherwise it adds complexity, tokens, and a security theater, without actually improving things.

Also, if we do want to implement it, I did so on another PR you can check out: https://github.com/ctxrs/ctx/pull/170

If we are going to do it, I think it should be like this.

Let me know what you think! I will close for now but we can keep discussing.

PR #163, opened 2026-07-13
Make provider import state converge safely

luca-ctx:
## Summary

- persist convergent provider import outcomes and parser revisions
- fence inventory, completion, and missing-path transitions with a durable per-source generation
- keep transient I/O failures retryable while deterministic record rejection converges
- repair legacy failed rows and completed rows whose materialized session is missing

## Why

Provider import state could previously remain ambiguous after mixed rejection, parser changes, or overlapping source scans. An older in-flight scan could also report against a newer observation. This change makes every durable result conditional on the exact current source observation without rewriting unchanged catalog rows.

## Validation

- `cargo test -p ctx-history-store catalog::tests` (34 passed)
- `cargo test -p ctx-history-store schema::tests` (29 passed)
- `cargo test -p ctx-history-capture codex` (44 passed, 1 ignored manual benchmark)
- `cargo test -p ctx-history-capture retained_existing_content`
- `cargo test -p ctx commands::import::native::tests` (8 passed)
- `cargo test -p ctx --test native_provider_rejections` (16 passed)
- `cargo clippy -p ctx-history-store -p ctx-history-capture -p ctx --all-targets -- -D warnings`
- two independent adversarial review passes; all findings resolved or assigned to the separate write-pressure PR as preexisting scope

This is the first PR in a stacked import-indexing resilience series. It is ready for CI and review, but will be held from merge until the combined stack passes cross-platform and real-corpus validation.


luca-ctx:

PR #165, opened 2026-07-14
Canonicalize test temp roots on macOS

### LINKED ISSUE #161: macOS: running the test suite can require manually canonicalizing temporary paths
While running the test suite on macOS I ran into an issue that took a while to track down, so I wanted to document it in case others hit the same thing.

Some tests that exercise provider imports can fail with an error like:
```
invalid provider transcript path ...
symlinked provider transcript path components are rejected

```
I found this was caused by the temporary directory path on macOS. By default, temporary directories are often created under /var/..., which resolves through the /var -> /private/var symlink. The production symlink checks are working as intended, but the test environment ends up passing a non-canonical path.

Using the canonical /private/... path (or otherwise resolving the temp directory before running those tests) made the failures disappear.

I haven’t dug into what the best fix would be, and it may simply be something to document or handle in shared test helpers if @luca-ctx think it’s worthwhile. I mainly wanted to report the contributor experience in case others run into the same thing.

luca-ctx:
## Summary

- canonicalize the system temporary root before creating provider-import test directories
- route provider-source discovery tests through the same canonical test helper
- preserve production symlink rejection unchanged

## Root cause

On macOS, temporary directories are commonly returned under `/var/folders/...`, while `/var` resolves to `/private/var`. Tests either passed the lexical `/var/...` path into production validation or compared it with a canonical `/private/var/...` discovery result.

## Validation

- reproduced the original failure on an Apple M1 running macOS 26.2
- `cargo test -p ctx-history-capture`: 259 passed, 1 ignored
- native-import unit tests: 5 passed
- native provider real-shape tests: 2 passed
- native provider rejection tests: 10 passed
- `cargo fmt --all --check`
- `git diff --check`
- independent implementation review completed with no findings

Closes #161.

### DIFF
--- a/crates/ctx-cli/src/commands/import/native_tests.rs
+++ b/crates/ctx-cli/src/commands/import/native_tests.rs
@@ -6,9 +6,18 @@ use ctx_history_core::{
 use ctx_history_store::{SourceImportFile, SourceImportFileIndexUpdate};
 use serde_json::json;
 
+fn tempdir() -> tempfile::TempDir {
+    let temp_root = fs::canonicalize(std::env::temp_dir())
+        .expect("system temporary directory should be canonicalizable");
+    tempfile::Builder::new()
+        .prefix("ctx-native-import-")
+        .tempdir_in(temp_root)
+        .unwrap()
+}
+
 #[test]
 fn codex_preinventory_failures_survive_when_catalog_has_no_pending_sessions() {
-    let temp = tempfile::tempdir().unwrap();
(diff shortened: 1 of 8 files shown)


luca-ctx:
The reviewed head was integrated into main by merge commit 89aa20d64826d69de62246bf9dee89e026e88a6e. GitHub's merge mutation raced with the base update and left this PR metadata open, so this already-integrated PR is being closed manually.

PR #166, opened 2026-07-14
Observe logical provider import units

luca-ctx:
## Summary

- inventory provider history as explicit logical import units instead of treating every source root as one opaque file
- observe SQLite databases with stable WAL/rollback-journal snapshots and companion dependencies
- persist source observations, unit outcomes, and inventory-generation completion behind concurrency fences
- retry superseded inventories without allowing stale work to report success

## Why

This gives incremental import a deterministic answer to “what changed?” across all provider adapters and prevents concurrent setup, search, or daemon work from committing observations for an obsolete source generation. It is the correctness layer required before append-only readers and bounded background scheduling can safely avoid full rescans.

## Validation

- `cargo test -p ctx -- --test-threads=1`
- `cargo test -p ctx-history-store -p ctx-history-capture`
- `cargo clippy -p ctx -p ctx-history-store -p ctx-history-capture --all-targets --all-features -- -D warnings`
- `cargo check -p ctx -p ctx-history-store -p ctx-history-capture --target x86_64-pc-windows-gnu`
- `cross check -p ctx -p ctx-history-store -p ctx-history-capture --target x86_64-unknown-freebsd`
- `cargo fmt --all -- --check`
- adversarial review completed with no remaining blockers

Stacked on #163; review only the top commit (`a5685aab`).

PR #169, opened 2026-07-14
Make provider file refresh append-safe and atomic

luca-ctx:
## Summary

- declare explicit append-only contracts for Codex session, Pi, Claude, and Tabnine JSONL sources
- persist bounded, validated append checkpoints and adapter resume state without re-reading the historical prefix
- publish replacement generations atomically with owner-scoped visibility fencing, bounded reconciliation, and convergent crash retirement
- preserve full-replacement output semantics except for revisioned Codex and Tabnine corrections

## Correctness

- incomplete trailing records never advance checkpoints
- shrink, replacement, identity mismatch, parser revision, or boundary mismatch rebuilds safely
- append and replacement writes share one owner lease and cannot expose mixed generations
- crashed mutated publications recover as replacement and can retire after their observation disappears
- streamed provider imports preserve whole-import source, session, edge, and archive metadata

## Validation

- 394 serialized ctx-history-capture tests passed; 1 manual performance test ignored
- 5 Cursor integration tests passed
- 175 ctx-history-store tests passed; 1 subprocess helper ignored
- strict all-target Clippy passed for capture and store
- Windows GNU and FreeBSD all-target cross-checks passed
- formatting and diff checks passed

This PR is stacked on #166. Native platform runtime smoke remains part of the cumulative stack gate.


luca-ctx:

PR #170, opened 2026-07-14
Harden agent-facing history output with response nonces

luca-ctx:
## Summary

- automatically frame default `ctx search` text, `ctx show session`/`event`
  text and Markdown, and table-formatted `ctx sql`; there is no opt-in flag
- wrap the complete agent-facing response with one fresh UUIDv4 nonce and a
  preamble that identifies the authoritative nonce; nested, mismatched, or
  replayed markers remain historical data
- automatically apply the same boundary to MCP search/SQL/show text and expose
  the same nonce as additive `structuredContent._ctx_history_envelope` metadata
- preserve machine-readable JSON/JSONL/CSV/raw contracts without framing and
  keep text search refresh diagnostics inside the response boundary
- update bundled skills, plugin instructions, CLI docs, MCP docs, and the threat
  model so agents use and interpret the boundary consistently

## Security boundary

This is provenance and delimiter-spoofing defense in depth. Retrieved history is
not escaped or rewritten. The boundary does not authenticate historical claims,
make recalled instructions safe, or replace current policy and authorization.

This is an independent, response-wide alternative design related to #160; it
does not reuse that PR's implementation.

## Validation

- `cargo test -p ctx --locked`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `bash scripts/check-docs.sh`
- `bash scripts/sync-plugin-skills.sh --check`
- isolated-HOME binary dogfood covering fixture import, search, all show
  surfaces, SQL, MCP, automatic default framing, exact machine formats, removal
  of the opt-in flag, unique nonces, and adversarial/replayed closing markers
- independent design review followed by an independent code review; all
  findings were fixed and the reviewer gave a clean final sign-off

### DIFF
--- a/crates/ctx-cli/src/agent_output.rs
+++ b/crates/ctx-cli/src/agent_output.rs
@@ -1,0 +1,96 @@
+use serde_json::{json, Value};
+use uuid::Uuid;
+
+const ENVELOPE_VERSION: u64 = 1;
+const TRUST_CLASSIFICATION: &str = "untrusted";
+const ENVELOPE_SCOPE: &str = "all other fields in this structuredContent object";
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct HistoryEnvelope {
+    nonce: Uuid,
+}
+
+impl HistoryEnvelope {
+    pub(crate) fn new() -> Self {
+        Self {
+            nonce: Uuid::new_v4(),
+        }
+    }
+
+    #[cfg(test)]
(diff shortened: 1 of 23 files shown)


luca-ctx:
Closing as through testing, it was determined that adding a nonce did not meaningfully help avoid prompt injection with GPT-5.6-sol. It may be possible that nonce could help, but I was not able to reproduce a scenario where it did help. If anyone wants to contribute to this, the best way would be to find a prompt injection attack that works if embedded in ctx history command results, but is mitigated if those results are wrapped in a nonce tag.

PR #172, opened 2026-07-16
Bound history refresh work and disk pressure

luca-ctx:
## Summary

- schedule fresh provider observations before bounded recovery work
- persist publication phases so daemon slices make durable, restart-safe progress
- detect same-stat ordinary-file rewrites with secure content observations
- pace provider reads, SQLite snapshots, store writes, and worker threads before physical I/O
- split the expanded import, search, capture, and store modules under the repository LOC policy

## Why

Initial setup and background refresh could issue sustained unpaced reads and writes, while recovery work could delay newly changed history. This layer makes refresh incremental, bounded, crash-convergent, and suitable for quiet daemon operation without weakening import freshness or publication atomicity.

## Validation

- final source tree is byte-for-byte equivalent to the previously proven cumulative implementation
- repository LOC and diff checks pass
- isolated GCP all-target check, strict Clippy, serialized full-suite, and cross-platform validation are being completed on the published head
- adversarial correctness, resource-pressure, and integration reviews are in progress

Stacked on #169. Do not merge independently of #163, #166, and #169.


luca-ctx:

PR #180, opened 2026-07-19
fix(history): preserve Codex source root across imports

j2h4u:
## Problem

Incremental Codex tail imports dropped the configured sessions root and could rewrite source provenance to the individual JSONL file. Full and incremental imports could therefore describe the same source differently, leaving catalog/source state inconsistent.

## Change

- preserve the configured `source_root` during tail imports
- make the non-fast full-import path use the same provenance semantics as the fast path
- keep `raw_source_path` anchored to the actual JSONL file
- cover both fast and non-fast full import followed by a tail import

## Validation

- `cargo clippy -p ctx-history-capture --all-targets -- -D warnings`
- `cargo test -p ctx-history-capture --lib` (262 passed, 1 ignored)
- `cargo test -p ctx-history-capture --test cursor_native` (5 passed)
- `cargo test -p ctx` (full package test suite)
- adversarial review with DeepSeek V4 Pro and GLM 5.2; no blocking correctness findings

### DIFF
--- a/crates/ctx-history-capture/src/provider/codex/session.rs
+++ b/crates/ctx-history-capture/src/provider/codex/session.rs
@@ -413,16 +413,12 @@ pub fn import_codex_session_jsonl(
     if options.fast_event_inserts {
         return import_codex_session_paths_fast(vec![path.to_path_buf()], store, options, 0);
     }
-    let source_path = options
-        .source_path
-        .clone()
-        .unwrap_or_else(|| path.to_path_buf());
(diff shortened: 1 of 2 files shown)


luca-ctx:
Closing as resolved by the landed ctx 0.26 ingestion work. Current main routes Codex full and tail imports through NativePath, preserves the configured source root separately from the JSONL source path, and includes equivalent full-then-tail regression coverage.

PR #182, opened 2026-07-20
Fix import progress and Claude root discovery

IlyaGulya:
## Summary
- keep per-source import updates from finalizing the overall progress bar
- discover existing `~/.claude-*` history roots and `CLAUDE_CONFIG_DIR` roots
- add coverage for `~/.claude-work` and an environment-configured root

## Root cause
Each imported source emitted a terminal `done` progress event, so multi-source imports printed repeated 100% progress bars.

## Validation
- `cargo fmt --check`
- full targeted Rust tests were blocked because the local Cargo cache lacks locked `anyhow 1.0.103`; online index refresh stalled.


IlyaGulya:
sorry, made this PR accidentally

PR #184, opened 2026-07-21
cli: improve diagnostics for missing SQL input files

atharva-patill:
## Summary

Improves the error shown when `ctx sql --file` is given a path that doesn't exist.

Other parts of the CLI already check for missing paths before attempting to open them and return a clear, command-specific error. This applies the same pattern to the SQL command, so users get a more direct message instead of an OS-level `NotFound` error.

## Changes

- Check that the SQL file exists before opening it.
- Return a clear error when the file is missing (`SQL file does not exist: <path>`).
- Add a regression test covering the missing-file case.

## Testing

- `cargo fmt --check`
- `search_show_locate_sql` (29/29 passing)
- `cli_public_help_docs` (15/15 passing)

### DIFF
--- a/crates/ctx-cli/src/commands/sql.rs
+++ b/crates/ctx-cli/src/commands/sql.rs
@@ -73,6 +73,12 @@ pub(crate) fn read_sql_input(args: &SqlArgs) -> Result<String> {
         }
         (Some(sql), None) => Ok(sql.clone()),
         (None, Some(path)) => {
+            if !path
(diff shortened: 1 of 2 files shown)


luca-ctx:
Thanks for the PR! I am going to close this one as I think it actually makes more sense here to just forward the original OS error like it is already doing. It still surfaces that the file doesn't exist, so this PR would only add a special case in the logic that reformats it.

PR #185, opened 2026-07-21
cli: surface search pagination status

atharva-patill:
## Summary

Improves the terminal search experience by indicating when additional results are available.

The search backend already reports whether more results exist, but the human-readable output stopped after the last displayed result with no indication that the search was incomplete.

## Changes

- Display `More results available.` when additional search results exist.
- Keep JSON output unchanged.
- Add an integration test covering the terminal output.

## After

```text
$ ctx search "history" --limit 1

1. ...

More results available.
```

## Testing

- `cargo fmt --check`
- Relevant test suite passes


luca-ctx:
Thanks for the PR! We have a big overhaul PR for 1.0 with #207 that incorporates this. Took a slightly different approach, but now we do a real `limit + 1` search and print "More results available" and remove the false pagination keys as we are not supporting pagination right now.

PR #186, opened 2026-07-24
feat: add diskless Turso import for shared history

moto-taka:
## Summary

- Add `ctx turso import` to import discovered native provider histories through a process-memory ctx Store, without creating a persistent local ctx SQLite index.
- Make remote ingestion idempotent and safe across Macs by canonicalizing provider session dedupe keys.
- Document the current Turso/libSQL-only event-projection scope and large-history limitations.

## Verification

- `cargo fmt --check`
- `cargo check --workspace`
- `cargo test -p ctx -- --test-threads=1`
- `cargo test -p ctx-history-store in_memory_store_uses_no_filesystem_database_path`
- `ctx turso import --help`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,6 +109,39 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


moto-taka:
Closing: this work is intended to remain in the author's fork for now.

Issue #176, opened 2026-07-17
Post-v0.24 payload normalization change breaks reimport of existing indexes (`ProviderEventConflict`)


hampsterx:
## Summary

Commit `13e69e6d` ("De-overload provider retention metadata", #168) changed the normalized
provider payload shape. For events with no provider-supplied hash, the dedupe-key hash is
derived from that payload, so reimporting an event that an older binary already indexed now
produces a different hash for the same `(source_id, provider_index)`.
`reject_provider_event_hash_conflict` rejects that, and on the default import path the error
aborts the entire run.

Result: an index built before `13e69e6d` cannot complete an import afterwards. Reads keep
working, so the index just silently stops taking new sessions.

Bisected below. This is on `main`, not a released tag.

## Bisect

Same index copy (schema v46, never migrated), same Claude source tree, two adjacent commits:

| Build | Contains `13e69e6d` | Result |
|---|---|---|
| `a45c3956` (Merge #167) | no | **exit 0**, `imported_events: 1715`, `skipped_events: 472` |
| `6a95db1c` (Merge #168) | yes | **exit 1**, `ProviderEventConflict` |

```text
Error: store error: provider event conflict for provider-source/<source-uuid> at index 0:
existing hash fnv1a64:40b74f86ac45a7b4, new hash fnv1a64:8c201597ab2ffbfd
```

Both runs used `--data-root <scratch>` against a byte copy of the same pre-existing index, so
the only variable is the commit.

## Mechanism

1. `claude_event` (and the other providers touched by `13e69e6d`) changed the emitted payload
   keys. In `crates/ctx-history-capture/src/provider/providers/claude.rs`:

   ```diff
   -            "text": text,
   -            "truncated": truncated,
   +            "text": retained_text.text,
   +            "text_retention": retained_text.retention.as_json(),
                "content_preview": provider_capped_json(...),
   -            "content_retention": retention.as_str(),
   ```

2. `import_provider_capture_line` falls back to hashing that payload when the provider does
   not supply its own hash (`crates/ctx-history-capture/src/provider/importer.rs`):

   ```rust
   let event_hash = event
       .provider_event_hash
       .clone()
       .unwrap_or(compute_payload_hash(&payload)?);
   ```

3. The dedupe key is `provider-source:<source_id>:<provider_index>:<payload_hash>`, and
   `reject_provider_event_hash_conflict_from_rows` errors when a stored row has the same
[...]

PR #207, opened 2026-07-30
feat!: make provider history authoritative with Tantivy and flat-F32 search

luca-ctx:
## Summary

This PR replaces ctx’s legacy canonical `work.sqlite` history Store with the storage and retrieval architecture for ctx 1.0.

Provider-native history files and databases now own transcript content. ctx’s Core history data plane consists of rebuildable projections tied to a verified source generation:

```text
provider history
  → stable identities and typed source locators
  → immutable Tantivy lexical generation
      ├─ exact source hydration for search, show, locate, and MCP
      ├─ relational.sqlite metadata projection
      ├─ optional flat-F32 semantic generation
      └─ optional generation-bound Pro materialization
```

The PR:

- Replaces `ctx-history-store` and `ctx-history-search` with dedicated source-backed index and relational projection crates.
- Converts native providers, custom JSONL, and history-source plugins to stable identities, typed locators, and exact provider hydration.
- Builds immutable Tantivy candidates privately, verifies their source manifest and policy identity, and publishes them atomically.
- Stores searchable text as indexed terms without retaining message bodies or display previews in Tantivy.
- Replaces the production `sqlite-vec` path with immutable, memory-mapped F32 vector segments.
- Adds generation-bound lexical, semantic, relational, and Pro catch-up with explicit stale, partial, and unavailable states.
- Moves refresh ownership to the persistent daemon, including provider discovery, event-driven refresh, semantic work, recovery, and query-service supervision.
- Rewires setup, import, search, show, locate, SQL, status, doctor, MCP, SDK contracts, and release construction around the new generation model.
- Starts a fresh 1.0 history epoch and removes the old Store only after a new Tantivy generation has been verified and activated.

## Why Tantivy replaced SQLite FTS

The matched lexical bakeoff used the same 998 source files, 5.44 GB of history, and 371,017 searchable events:

| Metric | SQLite contentless FTS5 | Tantivy |
| --- | ---: | ---: |
| Cold build wall time | 30.60 s | **7.65 s** |
| Main index size | 209.0 MB | **117.4 MB** |
| Total filesystem output | 3.87 GB | **240.8 MB** |

The SQLite writer was occupied for 30.587 of the 30.601 seconds, while 16 producers accumulated 304.4 seconds waiting to submit work. Removing copied bodies reduced storage, but it did not remove the single-writer bottleneck.

The query comparison showed the same split:

- Rare term: SQLite 0.885 ms, Tantivy 1.092 ms.
- Medium-frequency term: SQLite 28.83 ms, Tantivy 1.413 ms.
- Broad term: SQLite 115.23 ms, Tantivy 1.178 ms.

SQLite remained competitive for selective queries. Tantivy won decisively for cold construction, index size, write parallelism, and frequent-term retrieval—the workloads that dominate a large local history corpus.

These numbers compare the lexical data planes rather than claiming a fourfold improvement for every end-to-end command. The production implementation adds source verification, exact hydration, publication recovery, projections, and user-facing contracts on top of the measured engine.

## Why flat-F32 replaced sqlite-vec and ANN

The semantic bakeoff began from an existing embedding corpus and compared storage and query backends across 466,000 chunks:

| Backend | Build time | Query p50 | Storage | Recall |
| --- | ---: | ---: | ---: | ---: |
| sqlite-vec F32 | 5.83 s | 305 ms | 800 MB | exact baseline |
| USearch F32 | 116 s | 30 ms | 785 MB | 100% |
| USearch BF16 | 61 s | 17 ms | 427 MB | 99.45% |
| Flat mmap F32 | **1.10 s** | **21.7–24.2 ms** | **716 MB** | **100%** |

Flat F32 also matched sqlite-vec exactly on the retained real-E5 query set. At this corpus size, an exact scan under 25 ms left too little latency to justify HNSW’s graph construction, tuning, approximate recall, and roughly 916–1,796 CPU-seconds of build work.

Immutable flat segments fit ctx’s ownership model:

- vectors are written sequentially and published through a checksummed manifest;
- readers pin one generation and scan it through mmap;
- queries retain a bounded top-k without loading a database or graph;
- interrupted writes leave the previous generation readable;
- model or projection changes invalidate and rebuild derived data cleanly;
- macOS, Windows, and Linux builds no longer need to compile and load a SQLite vector extension.

Embedding generation still dominates cold semantic ingestion. Flat exact search therefore remains the correctness baseline until measured corpus size or query concurrency demonstrates a need for an ANN accelerator. A future ANN layer can be generated from the flat generation without becoming the sole copy.

## Where SQLite still fits

SQLite remains in the parts of ctx that benefit from relational or transactional storage:

- `relational.sqlite` provides disposable metadata and stable read-only `ctx_*` views for `ctx sql`;
- `usage.sqlite` stores bounded, content-free local usage aggregates;
- provider-owned SQLite histories remain supported and are read through verified snapshot paths.

SQLite no longer owns canonical transcript content, lexical retrieval, or semantic vectors. That boundary preserves SQL inspection without making one database responsible for source authority, full-text indexing, vector search, rendering, and migration.

## User-visible behavior

Search candidates come from the active Tantivy generation, while snippets and complete transcripts are hydrated from the exact provider record identified by that generation’s locator. A source that moved, changed, disappeared, or became unsupported produces a typed availability error. ctx does not silently return a stale database copy.

The daemon owns bounded refresh and projection catch-up. Foreground queries read committed generations instead of becoming history writers. Relational, semantic, and Pro consumers can catch up independently, but they must match the active Core generation before their results are admitted.

Semantic search remains opt-in. Lexical search stays available without a model, while hybrid search uses semantic evidence only when its flat-F32 generation has compatible, complete coverage.

## Migration

ctx 1.0 builds a fresh source-backed epoch from current provider history. It does not migrate, read, or fall back to the legacy canonical Store.

Old history artifacts are deleted only after the replacement Tantivy generation reopens and verifies successfully. Failed or interrupted publication retains the previous safe state and completes through the next fix-forward setup or refresh. If a provider source no longer exists, ctx reports it as unavailable instead of recovering content from an old copied row.

## Validation

Release qualification covered native Linux 1.0 artifact construction, dependency advisory and ABI checks, persistent-daemon lifecycle, exact no-op and append refresh behavior, fault recovery, and candidate smoke testing.


hampsterx:
Upgraded from a 0.25-era build to 1.0 off main today and rebuilt from scratch. Clean: ~21.4k sources, ~22.8k sessions, ~866k events, no migration drama.

The part that matters to me is that event identity no longer depends on ctx's own normalization output. That was the actual root of #176: any change to the normalized payload shape silently made every existing index un-reimportable, so a routine refactor was a breaking change for users who wouldn't notice until imports had been failing for a while. Source-authoritative generations remove the whole category, and "rebuild the projection" is a real recovery path in a way "repair the canonical store" never was.

What survived the rewrite is the failure granularity. #247 is the instance: one record with an unsupported body shape fails the entire source, and with ~19k files behind one source that took my Codex transcripts to zero while search kept serving hits from the sibling prompt-history source.

So the question: with generations bound to a verified source manifest, is source-level fatality now a deliberate invariant (a partially-indexed source can't be reconciled against its manifest, so it has to fail whole), or is per-record tolerance still the intent inside a verified generation, and #247 simply a gap?

PR #231, opened 2026-08-01
fix(capture): stop truncating Codex tool-call complete content

### LINKED ISSUE #224: `--content complete` silently truncates Codex tool-call arguments to preview length
## Problem

`ctx show event --content complete` can silently return a truncated Codex tool-call payload while reporting that the content is complete, source-verified, and not truncated.

This was reproduced with two real `function_call` records whose `arguments` values are valid JSON strings of 6,720 and 6,811 characters. The complete source JSONL records are 9,205 and 9,368 bytes.

For both events:

```bash
ctx show event <event-id> \
  --before 0 \
  --after 0 \
  --content complete \
  --quiet \
  --format json
```

returned event text of 4,016 and 4,015 characters. After removing the `<tool-name>: ` prefix, the retained argument text was exactly 4,000 and 3,999 characters. Both JSON objects ended in the middle of a string and were invalid.

At the same time, the response described the result as:

```json
{
  "complete": true,
  "complete_content_available": true,
  "origin": "provider_source",
  "source_verified": true,
  "stored_truncated": false
}
```

`ctx locate event` pointed to the correct source record and byte range, so the underlying provider data was present and readable.

## Likely cause

The current Codex NativePath tool-call projection builds display text through `codex_tool_arguments_preview` and then applies `PROVIDER_MAX_PREVIEW_CHARS` again in `build_tool_call`.

That is appropriate for indexed/display text, but the source-backed complete-content path appears to reuse the same preview projection. The exact provider record is successfully hydrated and verified, then its tool arguments are reduced to the 4,000-character preview before being returned as complete content.

## Expected behavior

- `--content complete` returns the full verified provider payload when it is within the complete-content byte budget.
- Preview limits used by the lexical/display projection do not apply to complete-content hydration.
- If any limit or retention policy is applied, the response must report truncation and must not claim `complete: true`.
- Structured tool arguments should not be silently converted into invalid partial JSON.

## Impact

Public-interface consumers cannot distinguish the result from genuinely complete content. In the observed case, identifiers near the end of the tool arguments were lost, so an analytics consumer had to fail closed and omit otherwise deterministic observations.

## Scope

This is separate from the aggregate hydration accounting problem in #218. The source record is small, hydration succeeds, and the response is returned; only the supposedly complete content is silently preview-truncated.

Environment: CTX 1.0.0 on Linux. Rechecked against current `main` at `b9da2c47`; no matching issue or pending fix was found.

rNoz:
## Summary

Fixes #224. `ctx show event --content complete` for a Codex tool-call event built its display text from `codex_tool_arguments_preview` / `codex_command_preview`, both capped at `PROVIDER_MAX_PREVIEW_CHARS` (4000 chars). That cap is correct for the short indexed/lexical summary but wrong for complete-content hydration: a large tool-call payload was silently cut mid-string while the response still reported `complete: true` and `stored_truncated: false`.

- split the shared file-touch-summarization / patch-field-redaction logic (previously inlined in `codex_tool_arguments_preview`) into `codex_tool_arguments_projection`
- add `codex_command_exact` and `codex_tool_arguments_exact`: uncapped counterparts of the existing preview functions
- `source_backed_tool_call_text`, the only caller feeding source-backed exact hydration, now builds its display text from the `*_exact` functions
- the preview-capped functions are unchanged and still back the `CodexToolCallContext` summary fields, so indexed/lexical body size is unaffected

## Validation

- new regression test: a tool-call with a `>4000`-char argument value, `hydrate_codex_locator` returns the untruncated text
- `cargo test -p ctx-history-capture --lib provider::codex::` (78 passed)
- `cargo fmt -p ctx-history-capture -- --check`
- `cargo check -p ctx-history-capture --lib`

Note: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test -p ctx-history-capture --lib` (full crate) currently fail on a clean checkout of `main` for reasons unrelated to this change (a `clippy::nonminimal_bool` lint on unrelated code in `ctx-history-index` and `ctx-history-capture`, and ~46 pre-existing macOS test failures around non-UTF-8 symlink handling and Gemini/Hermes/Zed fixtures). Verified these reproduce identically on bare `main`.


rNoz:
The bug I fixed here has landed upstream via the post-relational-architecture refactor (#245). Current crates/.../codex/nativepath/rows.rs:593-595 builds complete/exact content from the uncapped codex_command_text / codex_tool_arguments_text, while the PROVIDER_MAX_PREVIEW_CHARS-capped _preview variants are now used only for the lexical summary. 

That is exactly the split I proposed here. A waste of time to have this PR ready and fully tested, and not being considered on time... (it was ready 1d before your refactor)

PR #232, opened 2026-08-01
fix(capture): restore macOS symlink/socket rejection after data_root threading

### LINKED ISSUE #230: macOS: symlink-ancestor and socket rejection silently degrade to raw IO errors in root_handle
## Problem

On macOS (and other BSD-derived kernels), the root-authorized descriptor walk in `common/io/root_handle/unix.rs` silently weakens two of its rejection checks compared to Linux.

`open_component` opens each path component with `openat(parent, name, O_NOFOLLOW | O_DIRECTORY | ...)` and classifies the failure by `errno`:

- a symlinked ancestor is expected to fail with `ELOOP` and map to `AuthorityOpenError::Rejected`
- a non-regular special file (FIFO, device node, socket) is expected to fail with `ENXIO`/`ENODEV` and also map to `Rejected`

On macOS/BSD, `openat(..., O_NOFOLLOW | O_DIRECTORY)` against a symlinked ancestor returns `ENOTDIR` instead of `ELOOP`, and a bound Unix-domain socket returns `EOPNOTSUPP` instead of `ENXIO`/`ENODEV`. Neither errno is handled in `open_component`'s match, so both cases fall through to the catch-all arm and surface as a raw `std::io::Error` passthrough instead of the intended `Rejected` classification.

This was found while fixing two macOS-only tests in `complete_content/source_access/nanoclaw_snapshot.rs` that had been left on a stale call signature of `open_provider_sqlite_readonly`. The function gained a `data_root` parameter in `4f7ec06c` (fix(storage): secure selected roots and SQLite snapshots), but the macOS-gated tests were not updated, so `cargo test` does not catch this on Linux CI. Passing the missing argument to run the tests on macOS uncovered the errno-handling gap above.

## Impact

The symlink-ancestor and special-file rejection checks are part of the provider-source authority walk used to keep source access confined to trusted roots. On macOS, both checks degrade to a generic IO error rather than the specific `Rejected` outcome, which is what downstream resolver code (`provider_sources::resolvers::path_presence`, in one observed case) expects to classify the path correctly.

## Suggested fix

- add the missing `data_root` argument to the two stale macOS test calls
- add `EOPNOTSUPP` alongside `ENXIO`/`ENODEV` for the special-file case
- for `ENOTDIR`, disambiguate with an `fstatat(AT_SYMLINK_NOFOLLOW)` probe: reclassify to `Rejected` only when the component is confirmed to be a symlink, since `ENOTDIR` is also the legitimate error for a non-symlink component (e.g. a plain file standing in for a directory ancestor), which must keep surfacing as a raw IO error for existing callers

rNoz:
## Summary

Fixes #230.

- fix two macOS-only tests in `nanoclaw_snapshot.rs` that were left on the stale 1-arg signature of `open_provider_sqlite_readonly` after it gained a `data_root` parameter in `4f7ec06c`, which broke `cargo test` compilation whenever those `#[cfg(target_os = "macos")]` tests are included in the build
- passing the missing argument uncovered a real regression from the same commit's descriptor-walk changes: on macOS/BSD, `openat(..., O_NOFOLLOW | O_DIRECTORY)` against a symlinked ancestor fails with `ENOTDIR` rather than Linux's `ELOOP`, and a bound Unix-domain socket fails with `EOPNOTSUPP` rather than `ENXIO`/`ENODEV`; neither errno was handled, so both cases fell through to a raw `std::io::Error` instead of the intended `AuthorityOpenError::Rejected`
- disambiguate the `ENOTDIR` case with an `fstatat(AT_SYMLINK_NOFOLLOW)` probe before reclassifying, since the same errno is also returned for a legitimate non-symlink component (e.g. a plain file standing in for a directory ancestor) that must keep surfacing as a raw IO error for existing resolver callers
- update one resolver test whose expected diagnostic (`Unknown(NotADirectory)` for a path through a dangling symlink) is superseded by the more accurate `Unsupported` classification; the prior and new variants were already treated identically by `suppresses_fallback()`, so this is not a behavior change

## Validation

- `cargo test -p ctx-history-capture --lib common::io::` (23 passed)
- `cargo test -p ctx-history-capture --lib provider_sources::resolvers` (115 passed)
- `cargo test -p ctx-history-capture --lib nanoclaw` (21 passed, including the macOS-only symlink-ancestor regression test)
- `cargo fmt -p ctx-history-capture -- --check`
- `cargo check -p ctx-history-capture --lib`

Note: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test -p ctx-history-capture --lib` (full crate) currently fail on a clean checkout of `main` for reasons unrelated to this change (a `clippy::nonminimal_bool` lint on unrelated code, and ~46 pre-existing macOS test failures around non-UTF-8 symlink handling and Gemini/Hermes/Zed fixtures). Verified these reproduce identically on bare `main`.

### DIFF
--- a/crates/ctx-history-capture/src/common/io/root_handle.rs
+++ b/crates/ctx-history-capture/src/common/io/root_handle.rs
@@ -758,6 +758,42 @@ mod tests {
             .is_err());
     }
 
+    #[cfg(any(target_os = "linux", target_os = "macos", target_os = "freebsd"))]
+    #[test]
+    fn symlinked_ancestor_is_classified_as_a_rejected_provider_path() {
+        use std::os::unix::fs::symlink;
+
+        let temp = crate::test_support_paths::tempdir().unwrap();
+        let target = temp.path().join("target");
+        let linked = temp.path().join("linked");
+        fs::create_dir(&target).unwrap();
+        fs::write(target.join("source.jsonl"), b"inside\n").unwrap();
+        symlink(&target, &linked).unwrap();
+
+        let error = open_provider_source_file(&linked.join("source.jsonl")).unwrap_err();
+
+        assert!(matches!(
+            error,
+            CaptureError::InvalidProviderTranscriptPath { reason, .. }
+                if reason.contains("symlinked provider source path components")
+        ));
+    }
+
+    #[cfg(any(target_os = "linux", target_os = "macos", target_os = "freebsd"))]
+    #[test]
+    fn plain_file_ancestor_preserves_the_raw_not_a_directory_io_error() {
+        let temp = crate::test_support_paths::tempdir().unwrap();
+        let plain_file = temp.path().join("plain-file");
+        fs::write(&plain_file, b"ordinary").unwrap();
+
+        let error = open_provider_source_file(&plain_file.join("source.jsonl")).unwrap_err();
+
+        assert!(matches!(
+            error,
+            CaptureError::Io(error) if error.raw_os_error() == Some(libc::ENOTDIR)
+        ));
+    }
+
     #[test]
     fn descendants_reject_absolute_and_parent_escape() {
         let temp = crate::test_support_paths::tempdir().unwrap();
--- a/crates/ctx-history-capture/src/common/io/root_handle/unix.rs
+++ b/crates/ctx-history-capture/src/common/io/root_handle/unix.rs
@@ -272,9 +272,31 @@ fn open_component(
             Some(libc::ELOOP) => Err(AuthorityOpenError::Rejected(
                 "symlinked provider source path components are rejected",
             )),
-            Some(libc::ENXIO | libc::ENODEV) => Err(AuthorityOpenError::Rejected(
-                "provider source paths must be regular files or directories",
-            )),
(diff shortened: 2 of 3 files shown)


luca-ctx:
Maintainer update: I rebased this PR onto current main after merging #239, which separately reverted the duplicate implementation we had landed. The first commit remains rNoz-authored and contains the substantive fix from this PR. I dropped only the obsolete nanoclaw_snapshot.rs hunk because that file no longer exists after the intervening architecture changes, then stacked a separate maintainer commit that extracts the errno classifier and adds broader regression coverage.

Validated locally:
- cargo fmt -p ctx-history-capture -- --check
- cargo test -p ctx-history-capture common::io::root_handle (11 passed)
- cargo test -p ctx-history-capture dangling_links_and_symlink_loops_never_unlock_fallbacks (passed)
- cargo check -p ctx-history-capture --lib

I also attempted an aarch64-apple-darwin check, but this Linux host does not have an Apple C cross-compiler; bundled C dependencies stop on unsupported -arch/-mmacosx-version-min flags before checking the crate. PR #232 is now clean and mergeable.

PR #249, opened 2026-08-02
fix(cursor): skip non-regular source entries without failing discovery

### LINKED ISSUE #248: Cursor worker.sock socket blocks all provider history refresh
## Problem

`ctx setup --wait` starts the daemon but publishes no history or lexical generation, so `ctx search` returns nothing, even though Codex, Claude, OpenCode, Zed, Copilot CLI, and Auggie sources are all available.

The live Cursor tree contains `worker.sock`, a Unix-domain socket. The Cursor authority walk correctly refuses to open the socket, but the traversal treats every `open_child` failure as a fatal symlink rejection. That invalidates inventory completion, classifies the whole Cursor source as `unknown`, and the daemon then raises a terminal coverage failure for the entire multi-provider refresh. A single optional provider that happens to sit beside a socket takes down search for every provider.

## Reproduction

```
ctx sources --provider cursor --format=json
cat ~/.ctx/daemon/jobs/core-refresh.json
find ~/.cursor/projects ! -type d ! -type f -print
```

`ctx sources --provider cursor` reports Cursor as `unknown`, and the core refresh job records a terminal coverage error naming Cursor.

## Impact

Any user whose Cursor installation keeps a bound `worker.sock` under its projects tree loses all `ctx search` results, not only Cursor results, until the socket is removed. Removing or renaming the live socket is not an acceptable workaround.

## Suggested fix

Skip non-regular entries (Unix sockets, FIFOs, device nodes) during Cursor transcript traversal as a non-fatal issue instead of invalidating discovery completion. Keep the safe rejection behavior, and report the provider as available when its usable transcript files can be read. Genuine symlink and IO errors should remain fatal.

This is distinct from #230 (fixed), which corrected the low-level errno classification for the socket rejection on macOS but left the higher-level refresh still blocked by any single `unknown` optional-provider source.

rNoz:
## Summary

A live Cursor `worker.sock` Unix-domain socket sitting beside valid transcripts made `ctx setup --wait` publish no history or lexical index, so `ctx search` was unavailable for every provider even though Codex, Claude, OpenCode, Zed, Copilot CLI, and Auggie sources were present.

The Cursor authority walk opens each child with `open_child`. A non-regular entry (socket, FIFO, device node) is correctly refused with a `Rejected` error, but `visit_directories` treated every `open_child` error as a fatal symlink rejection. That invalidated inventory completion, classified the whole Cursor source as `unknown`, and made the daemon report a terminal coverage failure for the entire multi-provider refresh.

The refusal already carries a stable reason string. This change adds a shared `is_non_regular_source_rejection` classifier and skips those entries as a non-fatal `SpecialFile` issue, while keeping genuine symlink and IO errors fatal. Valid transcripts beside a special file now complete discovery, the provider reports available, and the full refresh publishes the lexical generation for all providers.

The safe rejection behavior is unchanged: the socket is still never opened or read, it is only no longer treated as coverage failure.

This is the higher-level counterpart to #230 (merged), which corrected the low-level socket errno classification on macOS but left the refresh still blocked by any single `unknown` optional-provider source.

## Validation

`cargo test -p ctx-history-capture --lib cursor_special_file_beside_transcript_is_skipped_without_failing_discovery` passes. The new regression test discovers a valid transcript alongside a non-regular special file and asserts completion is preserved, the transcript is still found, and the skipped entry is recorded as a non-fatal issue. It uses a FIFO to exercise the identical `NON_REGULAR_PROVIDER_SOURCE_REASON` rejection path without the platform `sun_path` length limit; the socket-specific errno mapping is covered by the existing `root_handle` unix tests.

`cargo clippy -p ctx-history-capture --all-targets` is clean for the changed surface.

Fixes #248.

### DIFF
--- a/crates/ctx-history-capture/src/common/io.rs
+++ b/crates/ctx-history-capture/src/common/io.rs
@@ -14,8 +14,8 @@ mod root_handle;
     reason = "provider adapters migrate to these capability types in follow-up slices"
 )]
 pub(crate) use root_handle::{
-    open_provider_source_file, open_provider_source_path, OpenedProviderSourceFile,
-    OpenedProviderSourcePath, ProviderSourceDirectory, ProviderSourceRoot,
+    is_non_regular_source_rejection, open_provider_source_file, open_provider_source_path,
+    OpenedProviderSourceFile, OpenedProviderSourcePath, ProviderSourceDirectory, ProviderSourceRoot,
 };
 
 /// Maximum directories admitted by one provider JSONL inventory.
--- a/crates/ctx-history-capture/src/common/io/root_handle.rs
+++ b/crates/ctx-history-capture/src/common/io/root_handle.rs
@@ -692,6 +692,24 @@ fn map_changed_open_error(path: &Path, error: AuthorityOpenError) -> CaptureErro
     }
 }
 
+/// Reason recorded when a provider source path component is neither a regular
+/// file nor a directory (for example a Unix-domain socket, FIFO, or device
+/// node). Traversal callers can skip such entries safely without treating the
+/// enclosing provider source as unreadable.
+pub(crate) const NON_REGULAR_PROVIDER_SOURCE_REASON: &str =
+    "provider source paths must be regular files or directories";
+
+/// True when `error` is the safe rejection of a non-regular special-file entry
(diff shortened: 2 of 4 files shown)


luca-ctx:
Thanks for the fix!

PR #263, opened 2026-08-03
fix(factory-ai-droid): retain retried records that reuse one native id

### LINKED ISSUE #261: factory-ai-droid import aborts with duplicate event identity when a tool call is cancelled and retried
## Problem

Factory AI Droid reuses a message `id` when a tool execution is cancelled and retried: the retry record keeps the same `id` and references it through `parentId`. The direct JSONL projector derives each event identity from the native record id alone, so both records mint the same stable event id inside one session and the whole import fails closed with `duplicate event identity <uuid> in one candidate generation`.

## Reproduction

```bash
work=$(mktemp -d)
mkdir -p "$work/sessions/project"
cat > "$work/sessions/project/demo.jsonl" <<'EOF'
{"type":"session_start","id":"demo","timestamp":"2026-07-14T09:30:00Z","cwd":"/workspace","model":"factory/droid"}
{"type":"message","id":"dup","timestamp":"2026-07-14T09:30:13Z","message":{"role":"user","content":[{"type":"tool_result","tool_use_id":"Execute_1","is_error":false,"content":"first"}]}}
{"type":"message","id":"dup","timestamp":"2026-07-14T09:30:23Z","message":{"role":"user","content":[{"type":"tool_result","tool_use_id":"Execute_2","is_error":true,"content":"Error: Tool execution cancelled by user"}],"visibility":"user_only"},"parentId":"dup"}
EOF
ctx import --provider factory-ai-droid --path "$work/sessions"
```

Got: the import aborts with `duplicate event identity ... in one candidate generation`.

Want: both records retained as distinct events.

## Impact

Any Droid session containing a cancelled and retried tool call makes the whole factory-ai-droid import fail, so no Droid history is imported at all. Real session logs hit this routinely, because cancelled executions are normal Droid usage.

## Suggested fix

Mint an occurrence suffix per (native record id, subrecord) pair, the same disambiguation the Cursor provider already uses. Occurrence zero must keep the exact current identity so previously imported events never move, and certified-suffix append scans must probe the base event identity lookup so appended retries continue the occurrence sequence instead of colliding with the imported prefix.

rNoz:
Fixes #261.

## Summary

- Factory AI Droid reuses a message `id` when a tool execution is cancelled and retried (the retry keeps the `id` and references it through `parentId`). The direct JSONL projector derived each event identity from the native record id alone, so the retry collided with the first occurrence and the whole import failed closed with `DuplicateEventIdentity`.
- Mints an occurrence suffix per (native record id, subrecord) pair, mirroring the disambiguation the Cursor provider already uses. Occurrence zero keeps the exact pre-fix identity, so previously imported events never move.
- Certified-suffix append scans probe the base event identity lookup (exponential then binary search) so appended retries continue the occurrence sequence instead of colliding with the imported prefix.

## Validation

- New unit test `factory_droid_reused_record_ids_mint_distinct_stable_event_identities`: two records sharing one native id mint distinct event ids, the first occurrence keeps the pre-fix identity, and replays are deterministic.
- New CLI end-to-end test `factory_droid_import_retains_retried_records_that_reuse_one_native_id`: importing a session with a reused id retains both records, and a certified-suffix append reusing the same id refreshes incrementally.
- `cargo test -p ctx-history-capture` and `cargo test -p ctx --test native_providers` failure sets are byte-identical to a clean `upstream/main` checkout (pre-existing environment-dependent sqlite and warp failures only).
- `cargo clippy -p ctx-history-capture -p ctx --all-targets` introduces no new warnings.
- Manual acceptance on macOS with a real `~/.factory/sessions` tree containing multiple sessions with reused record ids: `ctx import --provider factory-ai-droid` publishes successfully, cancelled-retry records are searchable, and a repeat import refreshes incrementally without identity errors.

### DIFF
--- a/crates/ctx-cli/tests/native_providers.rs
+++ b/crates/ctx-cli/tests/native_providers.rs
@@ -850,6 +850,107 @@ fn personal_agent_provider_imports_are_idempotent_and_incremental() {
 }
 
 #[test]
+fn factory_droid_import_retains_retried_records_that_reuse_one_native_id() {
+    // Factory AI Droid rewrites a message id when a tool execution is
+    // cancelled and retried, so one native record id repeats in a session.
+    let temp = tempdir();
+    let root = temp.path().join("native-droid-retry/sessions/project");
+    fs::create_dir_all(&root).unwrap();
+    let session_file = root.join("droid-retry.jsonl");
+    let shared_id = "droid-retry-shared-message";
+    let header = json!({
+        "type": "session_start",
+        "id": "droid-retry-session",
+        "timestamp": "2026-07-14T09:30:00Z",
+        "cwd": "/workspace",
+        "model": "factory/droid"
+    });
+    let result = |call: &str, is_error: bool, content: &str| {
+        json!({
+            "type": "message",
+            "id": shared_id,
+            "timestamp": "2026-07-14T09:30:13Z",
+            "message": {
+                "role": "user",
+                "content": [{
+                    "type": "tool_result",
+                    "tool_use_id": call,
+                    "is_error": is_error,
+                    "content": content
+                }],
+                "visibility": "user_only"
+            },
+            "parentId": shared_id
+        })
+    };
+    fs::write(
+        &session_file,
+        format!(
+            "{}\n{}\n{}\n",
+            header,
+            result("Execute_117", false, "factory-retry-initial-oracle"),
+            result("Execute_118", true, "factory-retry-cancelled-oracle"),
+        ),
+    )
+    .unwrap();
+    let path = root.parent().unwrap().display().to_string();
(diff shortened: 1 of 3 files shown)


luca-ctx:
Thanks again for the clear reproduction and the work on this!

During identity-stability testing, we found that occurrence-based suffixes can shift when an earlier duplicate is inserted or removed. That could cause an existing event ID (and therefore a citation) to resolve to different content.

We landed #265 using Factory’s validated parentId and tool_use_id linkage, which keeps retry identities stable across insertions, deletions, and reordering while continuing to reject ambiguous duplicates.

Next time, it would be great to stack this correction on top of your PR so we can keep your work.

PR #267, opened 2026-08-03
fix(kimi): index streamed assistant content parts from wire transcripts

rcny:
Kimi Code CLI never journals assistant replies as `context.append_message` — it persists them only as streamed `content.part` loop events, so the current classifier drops every model turn while keeping user prompts. This also affects tool events, which use `name` and `result.output` rather than the `toolName`/`output` shapes the importer expects.

- `content.part` now maps to a `Message` event with the `assistant` role, taking text from `part.text` and reasoning from `part.think`
- `/event/name` added to the tool-name fallback so modern `tool.call` records keep a body instead of being rejected as empty
- `/event/result/output` added to the output chain so `tool.result` payloads extract cleanly rather than serializing the wrapper object
- new unit tests pin the real wire-protocol 1.4 shapes for content parts and the tool exchange
- new e2e test in `native_provider_real_shapes.rs` asserts an event-scoped search finds the streamed reply; it returns 0 results before this change and 1 `assistant message` after

Existing fixtures use a schema the CLI does not emit, so session-scoped search masked this: the session still matched while the assistant event was absent from the event index.

### DIFF
--- a/crates/ctx-cli/tests/native_provider_real_shapes.rs
+++ b/crates/ctx-cli/tests/native_provider_real_shapes.rs
@@ -252,3 +252,54 @@ fn nanoclaw_import_preserves_text_timestamp_millis_and_integer_trigger() {
         "{search:#}"
     );
 }
+
+#[test]
+fn kimi_code_cli_wire_indexes_streamed_assistant_output_through_public_cli() {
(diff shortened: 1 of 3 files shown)


luca-ctx:
@rcny Thanks so much for the contribution! Great fix

PR #273, opened 2026-08-03
fix(kimi): re-project corrected wire semantics

rcny:
Follow-up to #267. That change corrected `content.part` classification but left `KIMI_SOURCE_PARSER_REVISION` at `v2`, so unchanged transcripts already indexed by the previous parser are never re-read.

This PR now uses one v3 projection boundary for both corrected Kimi wire semantics:

- bumps the revision to `kimi-code-cli-source-backed-v3`, forcing v2 transcripts to re-project cold
- normalizes the official Kimi `toolCallId` field into canonical `call_id` for both tool calls and results while retaining existing aliases
- keeps event identity stable because IDs derive from source, session, and ordinal rather than parser revision
- adds regression coverage proving both sides of a tool exchange publish `call_id: call_1`

Validation:

- `cargo clippy -p ctx-history-capture --lib -- -D warnings`
- `cargo test -p ctx-history-capture kimi -- --nocapture`
- `cargo test -p ctx --test native_provider_real_shapes kimi_code_cli_wire_indexes_streamed_assistant_output_through_public_cli -- --exact --nocapture`


rcny:
@luca-ctx Sorry to bug you again, but the previous PR was only part of the "full fix", this one handles Kimi transcripts that are already indexed.

luca-ctx (approved):
Thanks so much for the follow up - I also coupled this with a related fix to normalize kimi's tool call ID.

PR #279, opened 2026-08-03
Normalize the macOS system temp namespace safely

### LINKED ISSUE #277: tests fail on macOS with ENOTDIR under the default symlinked TMPDIR
## Problem

On macOS, running the test suite with the default per-user `TMPDIR` fails every test that calls `establish_private_data_root` with `Os { code: 20, kind: NotADirectory }` (28 failures in `cargo test -p ctx --bins source_backed_refresh_coordinator` alone, all at test setup).

macOS sets `TMPDIR` under `/var/folders/...`, and `/var` is a symlink to `/private/var`. `walk_private_directory` (`crates/ctx-history-core/src/platform_security/unix_private_directory.rs`) opens each path component with `openat(..., O_NOFOLLOW | O_DIRECTORY)`, so the symlinked `var` component is rejected and setup fails before the test body runs. `tempfile::tempdir()` follows `TMPDIR`, so every affected test inherits the symlinked root. `absolute_data_root_path` normalizes lexically but never canonicalizes, so the symlink survives into the walk.

## Impact

Any contributor running the suite on macOS sees dozens of failures unrelated to their change. It reads like a regression and costs review time; hit while validating #276.

## Reproduction

1. On macOS with the default environment: `cargo test -p ctx --bins source_backed_refresh_coordinator`.
2. Observe ENOTDIR panics at `establish_private_data_root` call sites.
3. `TMPDIR=/private/tmp cargo test -p ctx --bins source_backed_refresh_coordinator` passes 88/88.

## Suggested fix

Any of: canonicalize the data-root path in `absolute_data_root_path` before the walk (resolving `/var` to `/private/var`); have the affected tests canonicalize their tempdir; or at minimum fail with a clear "path contains a symlinked component" error instead of a bare ENOTDIR so the cause is obvious. The first option also helps real users who point `CTX_DATA_ROOT` at a symlinked path.

## Workaround

Run tests with a symlink-free temp dir: `TMPDIR=/private/tmp cargo test ...`.

luca-ctx:
Fixes #277.

macOS exposes the system temporary namespace through `/var`, while its canonical directory is `/private/var`. Normalize only that fixed platform alias before private-directory walking and data/source overlap checks.

All user-controlled descendants and final endpoints continue through descriptor-relative `O_NOFOLLOW` checks. Arbitrary symlinks—including aliases in user-owned `0700` directories—remain rejected, and non-macOS behavior is unchanged.

Validation:
- ctx-history-core: 70 tests passed
- source-backed refresh coordinator: 87 tests passed
- focused security review: passed
- aarch64-apple-darwin production/tests compile check: passed
- clippy, formatting, and diff checks: passed

Native macOS-specific runtime tests are included for CI/host execution.

### DIFF
--- a/crates/ctx-history-core/src/platform_security/path_overlap.rs
+++ b/crates/ctx-history-core/src/platform_security/path_overlap.rs
@@ -57,12 +57,13 @@ fn validate_provider_source_outside_data_root_with(
 ) -> io::Result<()> {
     let data_root = absolute_data_root_path(data_root)?;
     validate_absolute_path(source_root, "provider source root")?;
+    let source_root = normalize_platform_namespace_alias(source_root);
 
-    let source_before = inspect_named_endpoint(source_root, false)?;
+    let source_before = inspect_named_endpoint(&source_root, false)?;
     after_source_observation();
     let data = resolve_path(&data_root, true)?;
-    let source = resolve_path(source_root, false)?;
-    let source_after = inspect_named_endpoint(source_root, false)?;
+    let source = resolve_path(&source_root, false)?;
+    let source_after = inspect_named_endpoint(&source_root, false)?;
     if source_before != source_after {
         return Err(overlap_error(
             "provider source root changed during overlap validation",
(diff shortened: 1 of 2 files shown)


rNoz:
Thanks for the fix!

PR #291, opened 2026-08-04
Fix coordinated refresh and active-source snapshot integrity

### LINKED ISSUE #281: Manual import repeats a running cold daemon refresh instead of joining its result
## Problem

A manual all-provider import arriving while the daemon is already running a cold all-provider refresh is serialized as another complete pass instead of joining and reusing the in-flight result.

A fresh isolated reproduction on `504a9e4e` showed:

1. daemon startup admitted a periodic cold refresh;
2. `ctx import --all --no-daemon` arrived while that refresh was running;
3. the request reported `coalesced_requests: 0`;
4. ctx performed a second full provider pass and a second 1,020 MiB OpenCode online backup.

`--no-daemon` prevents autostart here; the explicit import still uses the already-running daemon endpoint.

Field evidence: https://github.com/ctxrs/ctx/issues/236#issuecomment-5171589000

## Current gap

The coordinator already has one active publisher, but its current attempt model conflates three things: the caller's logical freshness request, the physical executor pass, and terminal publication.

At current main, `ctx import --all` uses the automatic refresh request shape. A manual request admitted during an already-running cold refresh is conservatively forced into a successor chiefly by `fresh_after_admitted_snapshot`. Catalog `Some`/`None` mismatch is the analogous failure mode for explicit-path imports. In both cases, admission is decided from request metadata rather than semantic route coverage.

Relevant code:

https://github.com/ctxrs/ctx/blob/6312f7f85aa9acf3916d0234eee25ec268d37ed3/crates/ctx-cli/src/semantic/source_backed_refresh_coordinator/coordinator_state/request_lifecycle.rs#L338-L385

This is not an index-writer mutual-exclusion failure: only one publisher runs at a time. It is a single-flight coverage failure that repeats already completed work.

## Expected behavior

All refresh triggers should use one coordinator and one active publisher. A newer manual freshness request should attach to the running attempt. After publication, ctx should determine whether that result covers the requested catalog and admission-time freshness boundary:

- covered: reuse the same published generation and publication receipt, while recording the caller's own durable terminal request resolution;
- not covered: run only the uncovered or post-snapshot route delta;
- never repeat every provider merely because trigger metadata differs.

## Acceptance criteria

- A running periodic cold `All` refresh plus a concurrent explicit `import --all` launches one full executor and one large SQLite backup.
- The explicit caller receives its own durable terminal resolution, bound to the generation that satisfies its request.
- Provider writes after the running attempt snapshot create only a bounded delta successor.
- Duplicate waiters share terminal success and failure outcomes across daemon restart.
- No path weakens the single authoritative publisher or atomic generation publication contracts.

### LINKED ISSUE #282: Active provider writes can keep the daemon in an immediate refresh loop
## Problem

On a continuously active host, provider filesystem changes observed during a long refresh can immediately schedule more periodic work after publication. In the reported isolated run, ctx completed two expensive passes, published a partial generation, and then immediately started a third refresh with:

```text
trigger: periodic
trigger_provenance: daemon_scheduler
```

Field evidence: https://github.com/ctxrs/ctx/issues/236#issuecomment-5171589000

The prior reproduction observed the same behavior after a successful publication: a new full refresh began about 20 seconds later while provider histories were being written normally.

## Why this matters

Some follow-up work is legitimate when bytes were appended after a route snapshot. However, a long refresh should not turn normal continuous writes into an effectively permanent all-provider import loop. This repeatedly consumes CPU, I/O, scratch space, and large SQLite backup bandwidth.

## Expected behavior

Background maintenance should be generation- and route-aware:

- publication should satisfy watcher observations already covered by each certified route snapshot;
- only changes newer than that route snapshot should remain dirty;
- follow-up work should target exact changed routes rather than defaulting to another full pass;
- sustained writes should be debounced and rate-limited without disabling explicit freshness requests;
- retries should remain bounded and distinguish source-local failure from systemic failure.

This issue does not presuppose removing provider watchers. The product may retain background freshness, but it must have a bounded convergence and resource policy.

## Acceptance criteria

- Keep representative provider writers active throughout a cold all-provider refresh.
- After publication, ctx does not immediately start another full all-provider pass.
- A post-snapshot append schedules only its exact route after the documented debounce or cadence.
- Events already covered by the published route certificate do not recreate dirty work.
- Sustained writes cannot produce an unbounded tight refresh loop.
- An explicit wait/freshness request can still bypass background cadence and obtain a truthful request-bound result.

### LINKED ISSUE #283: Active Codex append can fail session-tree import during lineage scanning
## Problem

A normal append to an active Codex session can fail the Codex source-backed route with:

```text
provider source changed during bounded capture
```

The session-tree route otherwise has an append-tolerant contract: it certifies a frozen prefix and defers newer suffix bytes to the next refresh. The lineage helper currently reopens an ancestor session and requires exact file-observation equality before and after scanning it. An append during that scan is therefore promoted to a source failure.

Field evidence: https://github.com/ctxrs/ctx/issues/236#issuecomment-5171589000

Conflicting code paths:

- strict lineage scan: https://github.com/ctxrs/ctx/blob/6312f7f85aa9acf3916d0234eee25ec268d37ed3/crates/ctx-history-capture/src/provider/codex/nativepath/source_backed/lineage.rs#L116-L164
- append-tolerant route contract: https://github.com/ctxrs/ctx/blob/6312f7f85aa9acf3916d0234eee25ec268d37ed3/crates/ctx-history-capture/src/provider/source_backed/registration/families/jsonl/codex.rs#L376-L390

## Expected behavior

Each Codex source should have one admitted file identity and frozen EOF for a refresh. Ctx should parse and certify exactly that prefix, allow append-only growth beyond it, and derive lineage from the same captured authority rather than requiring a second quiescent read of the live file.

Rewrites, truncation, object replacement, and mutations within the certified prefix must still fail closed. Ordinary suffix growth must succeed and be caught up incrementally later.

## Suggested direction

Capture the tool-call and parent-session facts needed for lineage during the primary bounded parse, or spool those facts into bounded ctx-owned temporary storage. Resolve lineage from that generation-local captured data. A second read, if retained, must use the same frozen-prefix authority and append-tolerant prefix proof.

## Acceptance criteria

- Deterministically append to an active parent or ancestor session during lineage processing; the refresh publishes the certified prefix successfully.
- The next refresh imports the deferred suffix exactly once.
- Lineage results match a quiescent import of the same frozen bytes.
- In-prefix rewrite, truncation, source replacement, and invalid append proof remain source failures.
- Ambiguous or malformed lineage metadata is handled at the narrowest safe record/relationship boundary rather than failing an otherwise valid Codex source.

### LINKED ISSUE #284: OpenCode online backup can fail as malformed after copying the full database
## Problem

An active OpenCode source using the new logical online-backup path completed another approximately 1,020 MiB private copy and then failed as a source with:

```text
sqlite error: database disk image is malformed
```

The run preserved and published the healthy providers, so source-failure isolation worked. The OpenCode source remained unavailable after paying the full backup cost.

Field evidence: https://github.com/ctxrs/ctx/issues/236#issuecomment-5171589000

## What is known and unknown

The symptom is reproducible evidence of an unsuccessful OpenCode snapshot/import path, but the current diagnostic does not establish whether:

- the admitted live source view was already corrupt;
- the private online backup was inconsistent;
- a specific ordinary or virtual table was corrupt;
- the failure occurred during snapshot validation, schema observation, ordering, or row projection.

The implementation should not assume the backup API caused corruption until source and backup evidence distinguishes those cases.

## Expected behavior

A healthy WAL-active OpenCode database should yield a transactionally consistent private snapshot that remains readable for discovery, projection, and terminal certification while writers continue. If the provider source itself is corrupt, ctx should report that distinction with the failing phase and SQLite extended error while preserving other providers.

## Suggested investigation and direction

- Record the exact operation/phase and SQLite primary plus extended error code.
- In an isolated reproducer, run bounded integrity evidence against the admitted source view and private backup to identify where corruption first appears.
- Validate the private backup before beginning the expensive logical projection.
- Retain a safe opt-in diagnostic artifact or receipt sufficient to inspect a failed isolated backup without exposing provider content by default.
- If retry is appropriate, bound it and avoid immediately repeating a full 1 GiB copy indefinitely.

## Acceptance criteria

- Repeated backups of a healthy WAL-active OpenCode fixture remain valid while a writer commits concurrently.
- A source-corrupt database and a backup-corrupt database produce distinct phase-specific diagnostics.
- Projection uses one certified private snapshot from discovery through commit.
- Provider DB, WAL, SHM, and directory entries are not modified by ctx.
- Failure remains source-local, preserves the prior/readable generation, and cannot trigger an unbounded backup loop.

luca-ctx:
## Summary

This PR fixes four related active-source coordination and snapshot-integrity failures while retaining one authoritative publisher and atomic generation publication.

- **#281:** separates durable logical refresh demands from immutable physical attempts. A manual import can join a running cold refresh, evaluate that verified generation against its own admission fence, and schedule only an exact uncovered-route successor.
- **#282:** records monotonic route watermarks at watcher ingress and binds cleanup to exact verified-generation coverage. Automatic refreshes use bounded debounce, cooldown, and route-local retry instead of immediately rescanning active writers.
- **#283:** derives Codex lineage from the same admitted frozen-prefix scan that emits records. Ordinary suffix growth publishes the certified prefix and is imported once on the next refresh; lineage state stays bounded, task-local, disposable, and non-authoritative.
- **#284:** never opens the live provider SQLite family with SQLite. DB/WAL bytes are copied and certified through retained file authority into ctx-owned storage, the private logical backup is independently certified, and failures retain phase/artifact/code/cleanup provenance.

## Important boundaries

- There is still only one physical publisher. Explicit freshness can bypass background cadence, but it cannot create a second writer.
- Indeterminate, changed, or unavailable coverage fails closed to exact successor work; it is never treated as fresh.
- Codex lineage uses bounded in-memory facts and creates no durable sidecar or second source of truth.
- SQLite acceptance preserves the complete provider-directory name-to-bytes map, including DB, WAL, SHM, missing sidecars, and unrelated entries.
- SQLite resource exhaustion remains typed `ResourceUnavailable`; non-exhaustion `EIO` remains `Internal`.
- The removed dead `snapshot/online_backup.rs` path is not restored.

## Validation

- `cargo test -p ctx --bin ctx`: **1,462 passed, 3 ignored**
- `cargo test -p ctx-history-capture`: **1,070 passed, 1 ignored** across unit and integration targets
- Codex-focused capture: **172 passed, 1 ignored**
- SQLite-focused capture: **116 passed**
- `cargo clippy --workspace --all-targets -- -D warnings`
- `./scripts/check-loc.sh`
- `git diff --check origin/main...HEAD`

Validation was performed on Linux; this PR does not claim a Windows runtime receipt.

Closes #281
Closes #282
Closes #283
Closes #284


luca-ctx:
Focused follow-up validation on `1644dcdd7`:

- exact retained-overlay unit regression: PASS
- repeat Cline import + Cline→Roo carry/route-binding regression: PASS
- Codex mixed valid/malformed explicit receipt: PASS
- Pi mixed valid/malformed explicit receipt: PASS
- `cargo fmt --all -- --check`: PASS
- `./scripts/check-loc.sh`: PASS
- `git diff --check`: PASS

The exact tests prove deduplication uses provider + certified format + exact path (not lineage), repeated identical explicit imports remain no-ops, same-lineage relocation stays distinct, other explicit overlays remain carried, and partial explicit imports report exact source/rejection totals.

A fresh full native-provider target was also attempted; all owned cases passed before an unrelated later test hit host ENOSPC. The PR's previously recorded full-suite receipts remain unchanged, and the final changed contracts are covered by the focused reruns above.

PR #608, opened 2026-08-18
fix(opencode): ignore empty derived strings

j2h4u:
## Problem

A valid OpenCode SQLite source can contain empty derived metadata, for example a failed
tool call whose `state.input.command` is `""`. The OpenCode projection emitted that value
as a `ProviderDeclaredFact`, Core validation rejected the empty field, and the error escaped
the record projection. The result was that the complete OpenCode source was reported as
unreadable and none of its otherwise valid history became searchable.

## Change

- Skip exactly empty OpenCode fact values while preserving whitespace-only values, matching
  Core's existing `is_empty()` contract.
- Treat empty call-id, tool, status, and output-role strings as absent derived metadata.
- Preserve the complete structured source body and invocation arguments, including an empty
  command; only the invalid derived field is omitted.
- Bump the OpenCode parser revision to v10 so existing indexes are reprojected consistently.

## Validation

- Targeted Rust checks passed in the Microsoft Rust DevContainer.
- A release build imported an unchanged 1.5 GB OpenCode SQLite source into a fresh data root:
  168,782 searchable events across 1,504 sessions, with zero source failures.
- `ctx doctor` returned clean and provider-scoped lexical search returned OpenCode results.
- A production `import --all` then completed with Codex, Claude, OpenCode, Gemini, and Hermes
  present: 1,251,793 searchable events across 5,797 sources.
- The final diff received GO verdicts from two independent reviewers and a separate premium
  adversarial review.

### DIFF
--- a/crates/ctx-history-providers-sqlite-logical/src/providers/opencode/native_path/json.rs
+++ b/crates/ctx-history-providers-sqlite-logical/src/providers/opencode/native_path/json.rs
@@ -193,6 +193,21 @@ mod tests {
         };
         assert_eq!(diagnostic.body["exit_code"], 7);
         assert_eq!(diagnostic.body["command"], "false");
+
+        let (empty_role, _) = project(
+            r#"{"type":"tool_result","role":"","output":"ok"}"#,
+            "result",
+            None,
+            OpenCodeNativeSchemaFamily::SessionMessageSeq,
+        );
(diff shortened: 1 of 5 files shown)


luca-ctx:
Thanks for the fix! Merging

This exposes a deeper repeated issue I will handle as a follow-up, which is that event-level errors are bubbling up to become source-level errors and spoiling the entire provider source.

PR #649, opened 2026-08-21
fix(native-jsonl): reject duplicate records without aborting import

### LINKED ISSUE #648: factory-ai-droid import still fails closed when one message id is emitted under two different parents
Follow-up to ctxrs/ctx#261 (closed COMPLETED by #265).

## Problem

#265 stabilized self-parented retry identities (`parentId == id` with validated `tool_use_id` linkage) and kept every other duplicate shape fail-closed. Real Factory Droid sessions contain a shape that still aborts the whole import with `duplicate event identity <uuid> in one candidate generation`: one `message` id emitted twice, where neither copy is self-parented and the two copies hang under two different parents.

Observed in a real `~/.factory/sessions` tree: 3 of 777 session files, each with exactly one duplicated id. Example shape (ids anonymized):

```json
{"type":"message","id":"M","parentId":"P1","timestamp":"...","message":{"role":"user","content":[{"type":"tool_result","tool_use_id":"call_A","content":"..."}]}}
{"type":"message","id":"M","parentId":"P2","timestamp":"...","message":{"role":"user","content":[{"type":"tool_result","tool_use_id":"call_A","content":"..."}]}}
```

Both copies carry identical `tool_use_id` linkage, so the #265 discriminator cannot separate them, and their differing `parentId` is currently treated as contradictory evidence rather than as the discriminator. Because one ambiguous record id fails the whole candidate generation, a single such file makes every provider's history unavailable, not just Factory's.

## Impact

Any user whose Factory history contains this double-write shape permanently loses all ctx search until the file is removed. The shape arises from normal Droid usage (the affected sessions included cancelled and retried tool executions), so it recurs over time.

## Suggested fix

Treat native parent linkage as a discriminator for repeated record ids: the first in-file occurrence keeps the base event identity, and later occurrences with a distinct native `parentId` mint a `factory-ai-droid.repeated-record` subrecord selector from that parent id. Identities are therefore scan-order dependent: insertions, appends, and tail-scan resumes stay stable; deleting or reordering the first occurrence makes a later occurrence assume the base identity. Repeats without parent linkage, exact duplicates with identical linkage, and every non-Factory native-JSONL provider stay fail-closed.

Known residual: if the first in-file copy is later deleted, the survivor moves to the base identity, so a citation to its discriminated identity breaks. This is the same class of instability flagged on #263, bounded to the delete-first-copy case; making occurrence zero also parent-discriminated would avoid it at the cost of moving every existing Factory event identity once.

## Validation

- Unit: `factory_repeated_record_ids_are_retained_with_parent_linkage_discriminator` (distinct identities, first occurrence keeps the base identity, deterministic replay).
- CLI end-to-end: `factory_droid_repeated_record_ids_are_retained` (cold import retains both copies, no-op reimport is stable, a third emission arriving via certified-suffix append joins instead of colliding).
- Existing #265 regressions unchanged: `factory_droid_retry_lifecycle_keeps_retry_ids_stable`, `factory_retry_identity_is_source_scoped`, ambiguity and invalid-evidence fail-closed tests, and non-Factory strict rejection all pass.
- Real-data acceptance: `ctx import --all` over the tree that previously failed now publishes a verified generation (1,104 sources, 989 sessions, 168,234 events) and `ctx search` is ready.

rNoz:
Fixes #648.

Factory Droid can write the same physical message identity more than once in one session. When both rows projected to the same final ctx event identity, the index uniqueness safeguard aborted the entire direct-JSONL import, so valid records from that session never became searchable.

The final implementation keeps the failure local instead of inventing Droid-specific identities:

- Final event identities are checked per logical direct-JSONL source after normal provider projection.
- The first accepted physical record wins. If a later physical record would emit an identity already accepted from that source, the entire later record is rejected atomically and the import continues.
- Certified appends also check the published base, so an appended duplicate cannot collide during publication.
- Rejections flow through the existing bounded diagnostics, including provider, path, line, class, and detail.
- The policy applies to all direct-JSONL providers. Existing provider-native discriminators, including Factory's validated self-parented retry `tool_use_id`, continue to admit genuinely distinct events.

This supersedes the earlier parent-linked synthetic identity approach while preserving the contributor's original commits and authentic minimized fixture in the PR history.

Coverage includes the authentic #648 Factory fixture, cold/no-op/append behavior, atomic rejection of multi-event rows, first-record retention, diagnostic forwarding, the existing Factory retry lifecycle, source scoping, and a non-Factory Qoder collision.

Local validation on exact head `e0628ad7d524ecd83993e856ec91c661f0eca827`:

- `scripts/check.sh --mode ci`: passed
- all 435 CI build targets with Clippy warnings denied: passed
- all 207 CI test targets: passed
- crate-size, rustfmt, repository-policy, docs, SDK, and package gates: passed


luca-ctx:
Maintainer follow-up is stacked directly on this PR head and available at `ctxrs/ctx:ctx/issue-648-pr649-integration-20260821`, exact tip `97616cdd9889490787f3da1478c7c9521d654fb8`.

The stack preserves `50cbd75483eff1797ec3973d37b4df5cf8d24b74` as an ancestor, merges current `main`, and adds the stable parent-linked identity implementation plus upgrade-checkpoint migration correction. Please fast-forward this PR branch to that exact tip so the existing PR remains the integration path.

Local validation on the exact tip:
- required SDK checks: pass
- all 435 CI build targets with Clippy: pass
- no-cache CI tests: 204/207 pass
- affected Factory/native-provider unit and CLI contract suites: pass
- rustfmt, repository policy, and crate-size gates: pass

The three full-CI failures reproduce identically on pristine `origin/main` (`9812010b37d39dc2ad34f0ab4dd123bf6157b39c`) and are unrelated to this PR: `package_audit_fast`, `package_audit_release`, and `public_control_surface_check`, all concerning the Windows legacy-daemon takeover files.

Before merge, we also want one authentic pinned Factory capture of this exact double-write shape recorded as provider evidence.

rNoz:
All done for this PR, check PR desc + code.

Now, question on scope for a potential follow-up:

While preparing the authentic fixture I audited my own `~/.factory/sessions/` tree. The double-write shape this PR now handles (same native `message` id under two different `parentId` values with identical `tool_use_id` evidence) appears in 1 of 3 observed duplicate-message cases. The remaining cases are a different shape: a self-parented or differently-parented re-execution/cancellation where the duplicate copy uses a different set of `tool_use_id`s (e.g., a batch cancellation and a follow-up retry), which the current implementation still rejects.

Before I open a separate issue/PR for those re-execution cases, I wanted to ask: is that additional relaxation something you would like upstream, or do you prefer to keep the fail-closed surface narrow and scoped to the exact parent-linked identical-call shape covered here?

luca-ctx:
Maintainer update: I pushed the final implementation directly to this PR head at `e0628ad7d524ecd83993e856ec91c661f0eca827`. This is a normal fast-forward and preserves the contributor's original implementation commit, the authentic evidence fixture commit, and the intervening review commits in history.

After stepping back from the Droid-specific parent-linkage approach, the final tree removes that synthetic identity machinery. Direct-JSONL imports now keep the first physical record whose projected event identities are unique, atomically reject and report a later physical record that would reuse an accepted identity, and continue importing the rest of the source. The same check covers certified appends against the published base and applies to every direct-JSONL provider.

This also answers the follow-up scope question: we should not add more Factory parent/re-execution heuristics. Factory's existing validated self-parented retry discriminator (`tool_use_id`) still distinguishes legitimate retries. Other unexpected final-identity collisions are isolated as provider-record rejections rather than aborting the whole session or minting identities from reverse-engineered structure.

The authentic #648 fixture now completes with one reported rejection and five retained records; a no-op reimport preserves that result. A separate append contract proves that an appended duplicate is rejected against the published base. Generic Qoder and multi-event atomicity tests cover the provider-independent policy.

Local `scripts/check.sh --mode ci` passes on the exact head: all 435 lint/build targets and all 207 CI test targets are green.

PR #663, opened 2026-08-23
fix(claude): disambiguate repeated native record ids

msokh:
## Problem

Claude Code can re-emit a JSONL record that reuses an earlier `uuid`. The Claude adapter treated the native record id as unique, so two rows sharing a `uuid` derived the same event identity, and the candidate-generation audit rejected the generation:

```
duplicate event identity 137797a8-9a1c-8b8d-84b6-f4e0b88d5399 in one candidate generation
```

`failure_scope` for Claude is the whole `~/.claude/projects` route, so one such file blocked every Claude session from being indexed, and because the generation is shared, the periodic refresh published nothing at all (`published_generation == previous_generation`).

Observed trigger in real data: a `hook_success` attachment (`PreToolUse:ToolSearch`) re-emitted under the same `uuid`, the second copy carrying one extra field (`slug`). 11 such pairs in a single 1.8k-line transcript.

## Cause

`next_fallback_event_identity` returned `None` as soon as `row.native_record_id.is_some()`, so occurrence disambiguation only ever applied to rows *without* a native record id. The sibling adapters already do this for every row — Codex via `event_identity_for_occurrence`, Cursor via `next_event_occurrence` — which is why neither hits this.

## Change

- Resolve an occurrence for every row and carry it in the event key. Grouping key is the logical identity being repeated: `(native record id, subrecord index)` for native rows, the existing content digest for fallback rows, each domain-separated.
- `native_item_key` and `native_event_typed_key` shared the same parts already, so they now share one `event_key_parts`.
- Override `event_identity_revision` (`claude-native-record-occurrence-v1`) so existing bases are replaced in place. `SOURCE_SCHEMA_VARIANT` is deliberately left at v6: bumping it would change `SourceKey`, so an existing base would stop being the same source rather than being replaced.

## Tests

Three cases in `crates/ctx-history-capture-composition/src/source_backed/tests/claude_cursor.rs`:

| case | pre-fix | post-fix |
|---|---|---|
| re-emitted `attachment` under one `uuid` (observed shape) | `DuplicateEventIdentity` | 2 distinct events |
| repeated `uuid` on message rows | `DuplicateEventIdentity` | 2 distinct events, both bodies retained |
| byte-identical repeats with no native record id | passes | passes (fallback regression guard) |

Full-suite state: `ctx-history-provider-claude-cursor` 18/18; `ctx-history-capture-composition` 184 pass with 3 failures (`hermes` x2, `inventory` x1) that reproduce identically on an unmodified tree here.

Real-data check: a 117-file `~/.claude/projects` tree that fails on v1.0.2 imports clean (`failed_sources=0`, `rejected_records=0`, 9,270 records), and the duplicated `uuid` lands as two events with native keys `[uuid, 0, 0]` and `[uuid, 0, 1]` — neither dropped, no repeated event id in the session.

## Not addressed

Claude's `failure_scope` is still the whole route, so any single fatal per-file error keeps taking down every Claude session. Codex quarantines per file and continues. Happy to open that separately.


msokh:
Concrete evidence for why this uses `event_identity_revision` rather than a `SOURCE_SCHEMA_VARIANT` bump, in case it is useful for review.

An earlier revision of this branch bumped the variant v6 -> v7. Publishing with it and then returning to v6 leaves the refresh unable to reconcile:

```
source-backed scan failed for claude: Internal: session UUID collision at
10bf9cd0-cd0c-8988-b6d3-3147d8d31f07: existing digest 10bf9cd0…c5f9, new digest 10bf9cd0…c5f9
```

`ctx locate session 10bf9cd0` shows the session is owned by a source whose schema is `claude-nativepath-jsonl-v7`, while the current scan derives the same session UUID under the v6 source. Session identity does not depend on the schema variant, but source identity does, so a variant change makes the old source keep ownership of session UUIDs the new source also claims, and the whole route fails before the unobserved source is dropped.

So the variant bump is not merely wasteful churn - it breaks an existing index, and symmetrically it would break a downgrade across any variant change. The identity revision avoids it entirely: `SourceKey` is unchanged, `FamilyCheckpoint::valid_for` rejects the stale checkpoint, and the source is replaced in place.

luca-ctx:
Hi @msokh thanks for this PR. Per my other comment in #664 , I solved the "catastrophic failure" separately.

For this thread, lets focus on the "duplicate Claude Code UUID" issue. I haven't been able to repro it myself and wonder if I'm using the right CC version or if its a CC bug or something.

Would you be able to provide more info about CC version that produced the duplicate UUIDs, maybe a santized example of the rollout, or even better, deterministic instructions to repro it.

luca-ctx:
In other words I'm trying to figure out, is this a desired deliberate behavior of CC to emit multiple duplicate UUIDs (in which case a counter or something could be warranted as it's part of their product behavior), or is it a weird bug or edge case from CC (in which case its best to just reject the 2nd and beyond observations of the same UUID we already saw and continue gracefully beyond that).

camjac251:
@luca-ctx in #664 you said you could not reproduce this because your corpus is mostly Codex, and asked whether the repeats are deliberate Claude Code behavior (favouring a counter) or a defect (favouring rejecting the second and later observations). My history is heavily Claude Code and reproduces it at scale, so here is what the records actually look like.

## What the duplicates are

I grouped every record under `~/.claude/projects` by repeated `uuid`. A few dozen transcripts were affected, giving roughly 17k duplicate groups. Within a group:

| shape | share |
| --- | --- |
| differ in a shared key's value | ~65% |
| byte-identical | ~25% |
| later copy adds keys, changes nothing | ~10% |

Groups are bounded at two or three records, never more. The record `type` never changes inside a group.

Keys added by the later copy are dominated by `slug`, the shape this PR reports. But the larger pattern is a shared key changing value:

```text
gitBranch   ~8300
promptId    ~4300
version     ~2800
parentUuid  ~1200
message      ~230
cwd          ~230
toolUseResult ~170
```

`message` changes in under 2% of groups. Payloads are almost always identical; what moves is ambient metadata.

## `version` says why

`version` differing inside one session file means the two copies were written by different Claude Code builds. Those transitions are always forward, usually adjacent releases:

```text
 852  2.1.84  -> 2.1.87 -> 2.1.87
 573  2.1.79  -> 2.1.80
 412  2.1.76  -> 2.1.77
  97  2.1.193 -> 2.1.197
```

and in every group I sampled, the `message` payload was byte-identical while `cwd` and `version` differed.

That reads as Claude Code rewriting earlier records when a session is resumed, restamping ambient context while preserving the original `uuid` and payload. The three-record groups are sessions resumed twice. It would also explain `slug` appearing only on the later copy: a field the newer build writes and the older one did not.

This is inference from record shapes. I have not confirmed intent with anyone on the Claude Code side, and someone who knows their persistence model may read it differently.

## Bearing on the decision

If the resume-rewrite reading holds, these are one logical event observed twice, not two events. That argues against a counter: distinct identities would add ~17k near-duplicate events from this corpus alone, nearly all with identical payloads, surfacing as repeated hits for one moment in a session.

But rejecting the second and later observations has its own costs:

1. The later copy is the fresher one. It carries the newer build's `version`, the current `cwd`, and sometimes a corrected `parentUuid`. Keeping the first preserves stale context.
2. About 2% of groups differ in real content, roughly 230 where `message` changed and 170 where `toolUseResult` changed. Rejecting later observations discards those.

There is also a third option already in effect. Reading the preflight added in #676, a transcript that repeats a stable event identity now fails its whole source:

```rust
"Claude transcript repeats a stable event identity at lines {} and {}"
```

That is a large improvement over failing every provider, and the line numbers are a real diagnostic win. But the unit dropped is the entire transcript. On my corpus that would silently remove a few dozen sessions from search, including several of the largest and longest-running ones, and nothing in normal output tells the user which sessions are missing or why. Users with hook-heavy setups appear to hit this far more often, so the exposure is uneven.

That seems worth weighing alongside the counter-versus-reject question: without a row-level answer, the file-level containment is what users get, and losing whole sessions is a bigger loss than either row-level option.

A last-wins merge on ambient fields, keeping one event per `(uuid, subrecord_index)`, fits the observed data better than any of the three. I have no view on where that belongs relative to this PR.

Happy to run further queries against this corpus if a particular shape would help, and to test a build against it once something lands.

camjac251:
One more thing from digging around this, and it affects whether this PR does anything once rebased.

#676 added `source_backed/preflight.rs`, which derives the event identity ahead of projection so a transcript that repeats one can be quarantined on its own instead of failing every provider. `stable_native_event_identity` builds the key inline:

```rust
NativeItemKey::composite(
    NATIVE_EVENT_KEY_NAMESPACE,
    vec![
        native_record_id,
        TypedKey::U64(row.identity.source_subrecord_index),
    ],
)
```

That is a third copy of the shape `event_key_parts` consolidates here, and it landed after this PR was written, so it is not somewhere the rebase will obviously point you.

As written, this PR gives the projection its occurrence part while the preflight keeps deriving the old two-part key. The preflight would still find a collision and still quarantine the whole transcript, so the fix would merge and the people hitting this would see no change.

I checked that rather than assuming it: adding an extra part to the projection key alone makes the two paths disagree, and the source is still rejected.

So it seems worth pointing `preflight.rs` at `event_key_parts` in the same rebase, and maybe pinning it, since nothing currently fails when the two drift apart. A test that derives one row's identity through both paths and asserts they match is enough to catch it.

I had opened #684 to remove the duplication on its own, but your consolidation covers the same ground and is broader, so I closed it rather than race this. The branch is still up if the regression test is worth lifting: https://github.com/camjac251/ctx/tree/shared-event-key

luca-ctx:
Thank you for adding this but even after testing it with deliberate CC version upgrades + resume old sessions that have hook calls, I still wasn't able to reproduce it.

The best thing that would be helpful would be a script or runbook that can deterministically reproduce the behavior.

camjac251:
Saw this went in. One note in case it is useful later, since you mentioned resume plus a version upgrade did not reproduce it for you.

Looking at the affected transcripts here, the repeats are not scattered through the file. They arrive as one contiguous replayed block: in a representative file the duplicate pairs sit at lines 2/1904, 3/1905, 5/1907, a constant offset, and the record at 1904 carries `parentUuid: null` exactly like line 2. A `type: "permission-mode"` record precedes each block.

So the conversation head gets rewritten into the same file under a second root. Every replayed record keeps its original `uuid`; the later copy adds `slug`, and in some files `version` and `cwd` have moved forward too.

That may be why resuming by hand did not do it. The trigger does not appear to be resumption on its own, but a resume that replays the head of the conversation into the same transcript as a fresh `parentUuid: null` root. Across this corpus that is roughly 17k duplicate groups in a few dozen files, always bounded at two or three copies.

I have a small script that builds this shape synthetically and reproduces it against a throwaway data root, if a regression fixture is ever wanted. Happy to post it or open it as a PR.

PR #684, opened 2026-08-25
fix(claude): derive preflight and projection event keys from one place

camjac251:
## Summary

- extract the native event key parts into one helper shared by the preflight audit and both projection key builders
- add a test asserting the preflight and the projection derive the same event identity for one row

No behavior change. All three call sites already produced identical keys.

## Why

#676 added `preflight::stable_native_event_identity`, which derives a Claude event identity so a transcript repeating one can be quarantined without failing every provider. It builds the key inline:

```rust
vec![
    native_record_id,
    TypedKey::U64(row.identity.source_subrecord_index),
]
```

`native_item_key` and `native_event_typed_key` build the same shape independently in `source_backed.rs`. That is three copies of one definition, with nothing failing if they drift.

They agree today, so this change is inert on its own. It matters for the next change to that key.

#663 adds occurrence disambiguation to the projection key so a repeated `uuid` stops colliding. That PR predates #676 and does not touch `preflight.rs`. Rebased onto main as written, the projection would stop producing duplicate identities while the preflight kept deriving the old two-part key, still found a collision, and still quarantined the whole transcript. The fix would merge and change nothing for the users hitting it.

Its description also notes that `native_item_key` and `native_event_typed_key` "shared the same parts already, so they now share one `event_key_parts`", which is the same consolidation as here; #676 simply added a third copy after that PR was written.

I confirmed the failure mode rather than assuming it, by temporarily giving the projection an extra key part and running the new test:

```text
assertion `left == right` failed: preflight quarantines a source on
identities the projection never produces
```

Reverted before committing.

## Testing

- `cargo test -p ctx-history-provider-claude-cursor --lib`, 35 passed
- `cargo clippy -p ctx-history-provider-claude-cursor --all-targets`, no warnings in this crate

I have not run the Bazel gates. `scripts/bazelw` pulls a pinned toolchain that is not installed here, so I stayed with cargo for the touched crate. Point me at the gate you want and I will run it.

### DIFF
--- a/crates/ctx-history-provider-claude-cursor/src/claude/nativepath/source_backed.rs
+++ b/crates/ctx-history-provider-claude-cursor/src/claude/nativepath/source_backed.rs
@@ -872,6 +872,15 @@ fn session_identity(source: &SourceKey, native_key: TypedKey) -> Result<StableEn
     .map_err(contract)
 }
 
+/// Key parts for a row that carries a native record id.
+///
+/// The preflight audit and the projection below must derive byte-identical
+/// event identities for the same row. Building the parts once is what stops the
+/// two from drifting apart.
+fn native_event_key_parts(native_record_id: TypedKey, subrecord_index: u64) -> Vec<TypedKey> {
+    vec![native_record_id, TypedKey::U64(subrecord_index)]
+}
+
 fn native_item_key(
     row: &ClaudeRetainedRow,
     fallback_identity: Option<FallbackEventIdentity>,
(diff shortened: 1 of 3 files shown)


camjac251:
Closing this in favor of #663.

That PR already consolidates `native_item_key` and `native_event_typed_key` into one `event_key_parts`, which covers the same ground as the helper here and is broader. The remaining piece, pointing `preflight.rs` at the shared parts so the audit and the projection cannot drift, belongs in that change rather than racing it. Details are on #663.

The branch stays up if the regression test is useful to lift.

PR #689, opened 2026-08-25
fix(search): explain removed --include-subagents option

createsomethingtoday:
## Summary

`ctx search` now includes primary and subagent sessions by default, but older skills and command templates can still pass the removed `--include-subagents` option. The generic unknown-argument error currently suggests the unrelated `--include-current-session` flag, which makes the removal look like missing subagent support.

This keeps the current all-agent/root-first search contract and adds a targeted human diagnostic that explains:

- subagents are already included by default;
- `--primary-only` is the explicit narrower scope;
- `ctx search --help` is the canonical recovery path.

The search docs also explain how to inspect and refresh a stale managed ctx skill after a CLI upgrade.

## Compatibility

This does not restore `--include-subagents` as a redundant no-op or change search results. The option remains rejected with a nonzero exit, and machine/JSON invocations retain the raw Clap error bytes. Only the interactive human diagnostic changes.

## Test evidence

- `cargo test -p ctx dispatch::parse::tests -- --nocapture` (10 passed)
- `bash scripts/check-docs.sh`
- `git diff --check`

The broader CLI unit target reaches 213 passing tests locally, then fails in six unrelated hosted core-capability tests. A representative failure (`legacy_refresh_request_still_writes_exactly_one_terminal_response`) reproduces unchanged from clean `origin/main` at `8c6d6708`.

### DIFF
--- a/crates/ctx-cli/src/dispatch/parse.rs
+++ b/crates/ctx-cli/src/dispatch/parse.rs
@@ -104,6 +104,20 @@ fn human_clap_document(
             )
         }
         ErrorKind::UnknownArgument
+            if leaf.as_deref() == Some("ctx search")
+                && arguments_select_option(arguments, "--include-subagents") =>
+        {
+            (
+                "The --include-subagents option has been removed".to_owned(),
+                Some(
+                    "ctx search already includes primary and subagent sessions. Use --primary-only only when you want to exclude subagent work."
+                        .to_owned(),
+                ),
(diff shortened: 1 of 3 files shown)


luca-ctx:
Thanks for the contribution however I think we won't apply this one.

The ctx skill actually already does have an updater. So if it is a managed skill, it should be tracked and automatically updated whenever the underlying ctx version is updated. If the skill has been customized, the updater won't apply so as to not overwrite the latest skill.

Perhaps there is an issue or bug with the skill updater?

PR #698, opened 2026-08-25
fix(integrations): show and refresh skills for unselected agents

HackAttack:
The unselected-target sweep recognized a skill only under the legacy ctx-agent-history-search name, so a native copy dropped out of both status and install once a migration renamed it to ctx. Cursor is never in the default selection because it also reads `.agents/skills`, so an explicit `--agent cursor` install first showed up as stale and then vanished, and later upgrades silently stopped refreshing it.

Recognize a skill at either the current or the legacy path. Preserving a local edit in a swept-in target is no longer fatal, because an explicit install of one agent should not fail over a copy nobody selected. Skip paths the installer refuses to write through, so a symlinked skill directory cannot fail the whole command either.


luca-ctx:
Thanks @HackAttack ! Much appreciated contribution

Issue #658, opened 2026-08-22
Semantic search dlopen failed on macOS 26.6.2 (Tahoe): ONNX Runtime dylib loads via dlopen but ctx reports dlopen failed


yusoofsh:
## Summary

Semantic search stays `pending` on macOS 26.6.2 (Tahoe, build 25G83, arm64) with ctx 1.0.2. The daemon reports `dlopen failed` for the installed ONNX Runtime dylib, even though the dylib is properly signed and loads fine from Python/`clang dlopen`. `codesign -v` passes.

This is not a duplicate of #522 (Tahoe CoreML publication pins) — searched `repo:ctxrs/ctx dlopen OR Tahoe OR 26.6` and only #522 came back, which is about publication pins, not dylib loading.

## Environment

```text
ctx 1.0.2 (managed install via https://ctx.rs/install, ctx.install.json present)
~/.local/bin/ctx  TeamIdentifier=6JJ2TTFB9A  hardened runtime (flags=0x10000)
macOS 26.6.2  Build 25G83  arm64 (Apple Silicon)
ONNX Runtime 1.27.0  macos-arm64
Runtime path: ~/.ctx/runtime/onnxruntime/1.27.0/macos-arm64/lib/libonnxruntime.dylib
Config: [search] semantic = true, [indexing] mode = "manual" (also repros with auto/full)
```

Runtime dylib after hosted install:

```text
Executable=~/.ctx/runtime/onnxruntime/1.27.0/macos-arm64/lib/libonnxruntime.dylib
Identifier=libonnxruntime  TeamIdentifier=SJSNARH4TG  Runtime Version=15.5.0
CodeDirectory v=20500 size=296954 flags=0x10000(runtime)  Signature size=9073
Timestamp=13 Jul 2026  codesign verify: 0  (no quarantine after xattr -c)
```

For reference, the `ctx` binary itself: `Identifier=ctx TeamIdentifier=6JJ2TTFB9A Runtime Version=15.5.0`

Before the hosted install, the same path had the Homebrew-managed runtime (adhoc, `flags=0x2(adhoc)`, no TeamIdentifier) — also failed with the same error. The tarball `https://cli.ctx.rs/storage/v1/object/public/releases/artifacts/stable/1.0.2/ctx-onnxruntime-macos-arm64.tar.gz` is correctly signed by SJSNARH4TG (the fix: `cp /tmp/onnx_inspect/lib/libonnxruntime.dylib` into place).

## Current behavior

```text
[...]

PR #722, opened 2026-08-26
fix(semantic): recover CoreML backfill and batch source pages

### LINKED ISSUE #721: Semantic backfill rejects previous CoreML cache and embeds per document
## Summary

On macOS arm64 with ctx 1.1.1, semantic search still cannot complete after locally bypassing the ONNX/library-validation failure described in #658. Two problems appear in the post-load path:

1. The current CoreML descriptor points to an artifact that returns HTTP 404 and rejects a complete cached bundle whose manifest was pinned by ctx before #522.
2. Source projection calls the embedder once per document, even though reconciliation already works one page at a time. This adds one embedder invocation per document during page reconciliation.

This is separate from the ONNX/library-validation failure tracked in #658.

## Reproduction

Environment:

```text
ctx 1.1.1
macOS 26.6.2 arm64
183,114 indexed documents
```

The current descriptor on `main` uses:

```text
artifact: https://cli.ctx.rs/storage/v1/object/public/releases/artifacts/stable/1.0.0/ctx-multilingual-e5-small-coreml-fp16-1.0.0.tar.xz
manifest: 20a94162aca7c2f9f65be27839cd6867ec1c54e142fdf0c652de20139dffbc19
```

The artifact currently returns HTTP 404. The local cache contains a fully validated bundle with manifest `576c68756563333fdf442e6859f2392ca0065b09a2cb5d73983e30de75df1ad6`, which was the manifest in ctx's CoreML contract immediately before #522. The current acquisition path ignores it because only the new manifest is accepted.

After allowing that exact historical manifest as a cache-only fallback, semantic indexing starts. Profiling the backfill then shows `reconcile_source_page` calling `embed_chunks` once for each source document instead of once for the page.

## Expected behavior

- A complete, validated cache entry for ctx's previous pinned CoreML manifest remains usable when the current publication is unavailable.
- Pending chunks from one source page are sent through one embedder call.
- Operators can temporarily raise the worker duty cycle for a large recovery without changing the default policy.

## Proposed fix

- Prefer the current production CoreML descriptor, then fall back only to the exact previous manifest already pinned in ctx history. Keep the existing bundle validation.
- Collect pending chunks for the page and call `embed_chunks` once.
- Add a bounded `CTX_SEMANTIC_ACTIVE_PERCENT` override (`1..=100`); leave the default at 25 percent.

I have a patch with focused tests and a live semantic query against the repaired 183,114-document index.

The fallback is cache-only. It does not repair the missing hosted artifact or release packaging.

Refs #658 and #522.

yusoofsh:
## Summary

- restore the missing pinned CoreML 1.0.0 archive at its immutable contract URL from an independently checksum-verified release-pipeline artifact
- embed all pending semantic chunks from a source page through one index-layer request
- keep model-sized batching, failed-batch-only retry, runtime lock release, ordering, and duty-cycle throttling inside the shared runtime
- preserve page-atomic source publication and frontier advancement

## Validation

- focused semantic index/model and daemon tests
- semantic release and dependency-boundary checks
- repository policy and complete private-disclosure gate
- scripts/bazel-affected.sh origin/main (86 targets)

Fixes #721.
Refs #658 and #522.


cmm:
this (first commit) makes semantic search usable instead of always-stale in practice, thank you

PR #726, opened 2026-08-26
feat(search): synchronize semantic index for manual wait

### LINKED ISSUE #725: Support daemon-free semantic search with on-demand synchronization
## Summary

Semantic search currently assumes a persistent ctx daemon for semantic indexing and query embedding. Please support an explicit daemon-free mode where semantic state is synchronized only when ctx is invoked for an authoritative refresh.

## Proposed contract

- `ctx setup --semantic --no-daemon` enables semantic search, persists manual indexing, and leaves no resident daemon.
- `ctx search --refresh wait` starts a bounded finite worker, refreshes Core, drains semantic catch-up for the published generation, and waits for semantic readiness before querying.
- The query path embeds locally when no daemon query endpoint is resident.
- `--refresh off` and `--refresh background` may query an existing compatible semantic generation locally, but never start a worker or run semantic indexing in manual mode.
- The finite worker exits after admitted Core and semantic work is terminal and query IPC is quiescent.
- Explicit semantic requests remain fail-closed: runtime, worker, or coverage failures return typed errors instead of silently becoming lexical searches.

The intended workflow is:

```bash
ctx setup --semantic --no-daemon
ctx search 'pricing decisions' --backend semantic --refresh wait
ctx search 'follow-up detail' --backend semantic --refresh off
```

This is complementary to #721 and #722, which repair CoreML cache acquisition and batch semantic backfill. On macOS, a strictly process-per-trigger design can still incur substantial CoreML/ANE model-load latency even with a cached compiled bundle; the lifecycle should expose that wait rather than reporting readiness early.

yusoofsh:
## Summary

- allow semantic opt-in with the existing manual indexing mode
- for direct CLI semantic or nonzero-weight hybrid `search --refresh wait`, publish and pin Core through the existing finite worker, fully reconcile semantic against that exact generation, and reuse one foreground model runtime for the query
- keep lexical and zero-weight hybrid requests, manual `background`/`off`, MCP, and the automatic daemon path unchanged

## Validation

- `scripts/check.sh --mode ci` (217/217 test targets passed)
- isolated source-built manual semantic `search --refresh wait` completed against the pinned Core generation and returned semantic results
- lexical and zero-semantic-weight smoke paths did not load the semantic model

Closes #725.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,13 @@ ctx setup --semantic
 ctx index
 ```
 
-Semantic search requires automatic indexing, which is the default. If you use manual indexing, run `ctx index mode auto` first. Lexical search remains available while embeddings build; once they are ready, hybrid search uses lexical and semantic evidence automatically.
+Automatic indexing is the default. For semantic search without a persistent
+daemon, run `ctx setup --semantic --no-daemon`, then use
+`ctx search <query> --refresh wait`; that invocation refreshes lexical and
+semantic data with a finite worker and waits for it to exit. Background and
+off refreshes remain process-free in manual mode. Lexical search remains
+available while embeddings build; once they are ready, hybrid search uses
+lexical and semantic evidence automatically.
 
 ctx does not send your prompts, transcripts, or indexed history to a cloud service, call model APIs, require API keys, or write into your source repositories. Transcript text is preserved rather than automatically redacted, so review copied output before sharing it outside your machine.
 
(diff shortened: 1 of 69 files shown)


yusoofsh:
Live macOS verification with the installed ad-hoc hardened release build (`adhoc,runtime`) passed:

- Config: manual indexing, semantic enabled, no resident daemon.
- `ctx search ... --backend semantic --refresh wait`: exit 0; freshness `completed`; requested/effective mode `semantic`; semantic status `ready`; Core and semantic generation `f20fd46b...`; 5 results.
- During the wait there was one parent search plus one bounded `--finite-core-worker`; after completion `pgrep -x ctx` was empty and daemon status was `stopped`.
- `ctx search ... --backend semantic --refresh off`: exit 0 in 5s; freshness `existing_generation`; requested/effective mode `semantic`; same generation; 3 results; no worker remained.

The authoritative wait took 338s on this machine because a fresh process still pays CoreML/ANE model-load latency even with the compiled bundle cached. The follow-up off query reused OS/model cache and completed in 5s.

PR #736, opened 2026-08-26
docs: document status JSON schema v2

atharva-patill:
Documented the existing v2 JSON schema used by status responses.

Added ctx status, ctx daemon status, and MCP status to the v2 exceptions and clarified the status schema version.


luca-ctx:
Thanks @atharva-patill !

Issue #766, opened 2026-08-27
Windows: every hosted install fails — the install transaction and install.ps1 disagree on the marker's install_path form, and the ctx Pro pair step rejects both forms


Plumvery:
## Environment

- Windows 11 Home 26200, x64, NTFS
- PowerShell 7.6.5 (`pwsh`)
- ctx **1.1.1** (stable), `install.ps1` fetched from `https://ctx.rs/install.ps1` on 2026-08-27 (104,785 bytes)
- Install dir: `C:\Users\ichik\.local\bin` (no reparse points, no case mismatch, owner `ichik`)

## Symptom

Every hosted install fails:

```
install.ps1: ctx could not complete its crash-recoverable hosted install transaction
```

This reproduces on a completely clean machine state (no `ctx.exe`, no marker, no `~/.ctx`, no leftover lock files).

Below are five distinct defects found in one install attempt. They are reported together because they are sequential blockers in the same code path: fixing any one alone still leaves the installer unable to complete.

---

## 1. The install transaction requires the extended-length `install_path`; `install.ps1` writes the plain form

`Invoke-HostedInstallTransaction` runs:

```
ctx.exe upgrade --hosted-transaction install \
  --install-path <path> --attempt-id <id> --marker-source <file> --binary-sha256 <sha>
```

The child's actual error (discarded by the installer — see §5):

```
Error: hosted install marker does not bind the signed executable and target
```

**A/B test.** Same staged binary, same `--attempt-id`, same `--binary-sha256`, same directory. The *only* difference is the form of `install_path` inside the marker source:

| marker `install_path` | `--install-path` | result |
| --- | --- | --- |
| `C:\Users\ichik\ctxtest2\ctx.exe` | plain | exit 1 — `does not bind the signed executable and target` |
| `\\?\C:\Users\ichik\ctxtest2\ctx.exe` | plain | **exit 0 — `"status": "committed"`** |
| `\\?\C:\Users\ichik\ctxtest2\ctx.exe` | `\\?\...` | **exit 0 — `"status": "committed"`** |
| `C:\Users\ichik\ctxtest2\ctx.exe` | `\\?\...` | exit 1 |

`install.ps1` writes the plain form:

```powershell
$marker = [ordered]@{
    ...
    install_path = $installPath          # C:\Users\ichik\.local\bin\ctx.exe
```

so as shipped this can never succeed on Windows.

[...]

PR #768, opened 2026-08-27
fix(upgrade): pass non-verbatim paths to the Windows extraction helper

### LINKED ISSUE #767: Windows: semantic runtime never installs — ctx runs its generated extract.ps1 without binding -Destination
## Environment

- Windows 11 Home 26200, x64
- ctx **1.1.1** (stable), managed install at `C:\Users\ichik\.local\bin\ctx.exe` (`upgrade status` → `managed: true`, `marker: valid`)
- Fresh data root (`~/.ctx` created by this install)
- ONNX Runtime sidecar artifact: `runtime-7fc0ed292117a84373268fd68e4c7f1dbdcb06fa5057c7ce44579b918dbe36c3.artifact`, 5,940,650 bytes, `VERSION_NUMBER` = `1.27.0`

## Symptom

The semantic runtime can never be installed. Any semantic-enabled upgrade/repair ends with:

```
Error: extract ONNX Runtime sidecar: Join-Path : Cannot bind argument to parameter 'drive' because it is null.
At C:\Users\ichik\.ctx\.ctx-upgrade-downloads\runtime-7fc0ed29...e36c3.extract.ps1:100 char:39
+ ... -Item -ItemType Directory -Path (Join-Path $Destination 'lib') -Force ...
    + CategoryInfo          : InvalidArgument: (:) [Join-Path], PSArgumentNullException
    + FullyQualifiedErrorId : ArgumentNull,Microsoft.PowerShell.Commands.JoinPathCommand
```

(The original is localized Japanese; the parameter named in the message is `drive`, i.e. `Join-Path` received `$null` for `-Path`.)

Through `install.ps1` the same failure surfaces as:

```
install.ps1: ctx Semantic runtime repair failed
```

## Cause: ctx invokes its generated `extract.ps1` without binding `-Destination`

The generated script declares four parameters:

```powershell
param(
  [string]$ArchivePath,
  [string]$Destination,
  [string]$ExpectedVersion,
  [long]$MaxExpandedBytes
)
```

**`$ArchivePath` and `$ExpectedVersion` are bound; `$Destination` is not.** That is provable from where it fails: the script opens the archive, enumerates entries, and passes the strict version gate at line 97

```powershell
if ($versionText -cne ($ExpectedVersion + [char]10)) {
    throw "runtime VERSION_NUMBER is not exactly $ExpectedVersion"
}
```

before dying at line 100 on `Join-Path $Destination 'lib'`. An unbound `[string]` parameter is `$null`, which is exactly what `Join-Path` reports. If the arguments were positional and one were missing, the version gate would have thrown first — it does not.

**The generated script itself is correct.** I captured it before ctx deletes it and ran it by hand against the same `.artifact`, with all four parameters supplied by name:

```powershell
& .\extract-captured.ps1 -ArchivePath <the .artifact> -Destination <empty dir> `
    -ExpectedVersion '1.27.0' -MaxExpandedBytes 400000000
```

It completes successfully and produces exactly the expected tree:

```
LICENSE
MICROSOFT_VC_RUNTIME_LICENSE.rtf
ThirdPartyNotices.txt
VERSION_NUMBER
GIT_COMMIT_ID
lib\onnxruntime.dll
lib\msvcp140.dll
lib\msvcp140_1.dll
lib\vcruntime140.dll
lib\vcruntime140_1.dll
```

So the defect is in the invocation, not in the script.

## Reproduction

```powershell
$env:CTX_SEARCH_SEMANTIC = '1'
ctx upgrade --channel stable --format=json
```

Reproduces with and without `CTX_RELEASE_METADATA_URL` / `CTX_RELEASE_METADATA_SIGNATURE_URL` pointed at a local metadata file (which is what `install.ps1`'s semantic repair step does), so it is not related to the metadata override.

Note that ctx deletes `<hash>.extract.ps1` after the failure while leaving `<hash>.artifact` behind, so the generated script has to be captured while it exists.

## Impact

Semantic search cannot be enabled on Windows at all on a clean install. It also aborts `install.ps1` at the semantic repair stage, so the install never reaches `ctx setup` or the skill install.

## Relation to #642

#642 was a different failure in this same path (the spawned `extract.ps1` could not open the `.download` file it was handed, because the downloader still held it). That specific failure is gone in 1.1.1 — the archive opens fine here. The extract path is still broken, now because of the unbound `-Destination`.

Happy to test a patch.

Plumvery:
Fixes the Windows semantic-runtime install failure reported in #767.

## What is happening

`extract_runtime_archive` drives the embedded ONNX Runtime extraction script through `System32\WindowsPowerShell\v1.0\powershell.exe`. The `-Destination` it passes is `staged_path` from `stage_downloaded_runtime_artifact`, which descends from `semantic_runtime_root(data_root)`. When the data root has been canonicalised — `std::fs::canonicalize` yields the verbatim `\\?\` form on Windows — every path below it inherits that prefix.

Windows PowerShell's provider cmdlets cannot resolve a drive for a verbatim path, so the script's first `Join-Path $Destination 'lib'` fails and the whole semantic runtime install aborts:

```
extract ONNX Runtime sidecar: Join-Path : Cannot bind argument to parameter "drive" because it is null.
At C:\Users\<u>\.ctx\.ctx-upgrade-downloads\runtime-<hash>.extract.ps1:100 char:39
+ ... -Item -ItemType Directory -Path (Join-Path $Destination 'lib') -Force ...
```

`-ArchivePath` survives the same treatment because the script consumes it through `[System.IO.Compression.ZipFile]::OpenRead`, which accepts verbatim paths. Only the *provider* cmdlets reject them. That is why `$ArchivePath` and `$ExpectedVersion` bind and work while `$Destination` looks unbound — the script gets far enough to open the archive and pass the `VERSION_NUMBER` gate before dying.

The user-visible effect is that `runtime/onnxruntime/<version>/` is created and `<platform>` never appears, so semantic search can never be provisioned on Windows.

## Evidence

Windows 11, `System32\WindowsPowerShell\v1.0\powershell.exe`, same machine and session:

```
Destination = C:\Users\<u>\AppData\Local\Temp\ctx-verbatim
  Join-Path $Destination 'lib'  ->  C:\Users\<u>\AppData\Local\Temp\ctx-verbatim\lib

Destination = \\?\C:\Users\<u>\AppData\Local\Temp\ctx-verbatim
  Join-Path $Destination 'lib'  ->  Cannot bind argument to parameter "drive" because it is null.
```

Feeding the real generated extractor a verbatim `-Destination` reproduces the production failure exactly, down to the same location (`extract.ps1:100 char:39`). The same script with an identical but non-verbatim destination extracts the expected tree (`lib/onnxruntime.dll`, `lib/msvcp140.dll`, …, `VERSION_NUMBER`, `LICENSE`).

## The change

One file. A `powershell_path_argument` helper that converts `\\?\C:\…` to `C:\…` and `\\?\UNC\server\share` to `\\server\share`, leaves every other path untouched, applied to the two path arguments handed to the helper.

## Tests

`cargo test -p ctx-upgrade-engine --lib` on Windows 11, cargo 1.98.0:

| | passed | failed |
| --- | --- | --- |
| before | 84 | 3 |
| after | 88 | 3 |

The four added tests cover verbatim drive paths, verbatim UNC paths, ordinary paths, and a non-drive device path that must be left alone.

The three failures are pre-existing on this platform and unrelated to this change — I confirmed by stashing the patch and re-running:

```
upgrade::install::transaction::tests::coreml_clean_install_journals_exact_fallback_composition_and_paths
upgrade::install::transaction::windows::tests::failed_replace_repairs_missing_executable_before_any_wait
upgrade::state::tests::only_trusted_automatic_active_state_admits_the_recovery_daemon
```

For what it is worth they are also canonical-path shaped — one reports `install transaction data root is not canonical: C:\Users\<u>\AppData\Local\Temp\.tmpXXXXXX\data-root`.

## End-to-end

Verified on Windows 11 with a release build of this branch installed into a throwaway `BinDir` with a valid hosted-install marker and a scratch `CTX_DATA_ROOT`. The extraction step completes and the staged runtime tree is written in full, including ctx's own `ctx-runtime-install.json`. Details, including one caveat about a later unrelated-looking check, are in a comment below.

## Related

This is the same underlying issue as the marker-path mismatch reported in #766: the internal canonical `\\?\` form reaching a consumer that cannot accept it. A single normalisation applied wherever a path leaves ctx for an external consumer would cover both, and possibly the three failing tests above. I kept this PR to the one defect.

## Optional follow-up

`scripts/test-windows-runtime-upgrade-extractor.ps1` supplies `-Destination` from the test itself, so it validates the embedded script text but cannot catch a caller that hands over an unusable value — which is exactly this bug. Adding an `Assert-EmbeddedExtractorSuccess` case with `-Destination "\\?\$runtimeDestination"` would fail on current `main`; making it pass would also mean hardening the embedded script to use `[System.IO.Path]::Combine` / `[System.IO.Directory]::CreateDirectory` instead of `Join-Path` / `New-Item`. Happy to add that here or as a separate PR, whichever you prefer.

Fixes #767

### DIFF
--- a/crates/ctx-upgrade-engine/src/upgrade/install.rs
+++ b/crates/ctx-upgrade-engine/src/upgrade/install.rs
@@ -466,9 +466,14 @@ try {
   if ($versionText -cne ($ExpectedVersion + [char]10)) {
     throw "runtime VERSION_NUMBER is not exactly $ExpectedVersion"
   }
-  New-Item -ItemType Directory -Path (Join-Path $Destination 'lib') -Force | Out-Null
+  [void][System.IO.Directory]::CreateDirectory(
+    [System.IO.Path]::Combine($Destination, 'lib')
+  )
   foreach ($name in $expectedFiles) {
-    $target = Join-Path $Destination ($name.Replace('/', '\'))
+    $target = [System.IO.Path]::Combine(
+      $Destination,
+      $name.Replace('/', [System.IO.Path]::DirectorySeparatorChar)
+    )
     $sourceStream = $entries[$name].Open()
     try {
       $targetStream = [System.IO.File]::Open($target, [System.IO.FileMode]::CreateNew, [System.IO.FileAccess]::Write, [System.IO.FileShare]::None)
--- a/crates/ctx-upgrade-engine/src/upgrade/install/archive.rs
+++ b/crates/ctx-upgrade-engine/src/upgrade/install/archive.rs
@@ -52,34 +52,6 @@ fn windows_powershell_path() -> Result<std::path::PathBuf> {
         .join("powershell.exe"))
 }
 
-/// Convert a path into the form Windows PowerShell can accept as a parameter.
-///
-/// Paths inside ctx are frequently canonicalised, which on Windows yields the
-/// verbatim `\\?\` form. The provider cmdlets the extraction helper relies on
-/// (`Join-Path`, `New-Item`) cannot resolve a drive for a verbatim path and fail
-/// with a null `drive` binding, so the ordinary Win32 form is passed instead.
-#[cfg(windows)]
-fn powershell_path_argument(path: &Path) -> std::ffi::OsString {
-    use std::ffi::OsString;
-
-    let Some(text) = path.to_str() else {
-        return path.as_os_str().to_owned();
-    };
-    let Some(rest) = text.strip_prefix(r"\\?\") else {
-        return path.as_os_str().to_owned();
-    };
-    if let Some(share) = rest.strip_prefix(r"UNC\") {
-        return OsString::from(format!(r"\\{share}"));
-    }
-    let mut characters = rest.chars();
-    let drive = characters.next();
-    let colon = characters.next();
-    if drive.is_some_and(|drive| drive.is_ascii_alphabetic()) && colon == Some(':') {
-        return OsString::from(rest.to_owned());
-    }
-    path.as_os_str().to_owned()
-}
-
 #[cfg(unix)]
 pub(super) fn extract_runtime_archive(
     _process: &dyn ReleaseProcessPort,
@@ -246,9 +218,9 @@ pub(super) fn extract_runtime_archive(
         .args(["-NoProfile", "-ExecutionPolicy", "Bypass", "-File"])
         .arg(&script_path)
         .arg("-ArchivePath")
-        .arg(powershell_path_argument(archive_path))
+        .arg(archive_path)
         .arg("-Destination")
-        .arg(powershell_path_argument(destination))
+        .arg(destination)
         .arg("-ExpectedVersion")
         .arg(version)
         .arg("-MaxExpandedBytes")
@@ -603,7 +575,7 @@ param(
 $ErrorActionPreference = 'Stop'
 Add-Type -AssemblyName System.IO.Compression
 Add-Type -AssemblyName System.IO.Compression.FileSystem
-$contract = Get-Content -LiteralPath $ContractPath -Raw | ConvertFrom-Json
+$contract = [System.IO.File]::ReadAllText($ContractPath) | ConvertFrom-Json
 $expectedArchive = @{}
 $directories = [System.Collections.Generic.HashSet[string]]::new([System.StringComparer]::Ordinal)
 foreach ($file in $contract.files) {
@@ -664,9 +636,12 @@ try {
     }
     $record = $expectedArchive[$name]
     if ([long]$entry.Length -ne [long]$record.size) { throw "Semantic zip file size mismatch: '$raw'" }
-    $target = Join-Path $Destination ([string]$record.path).Replace('/', '\')
-    $parent = Split-Path -Parent $target
-    New-Item -ItemType Directory -Path $parent -Force | Out-Null
+    $target = [System.IO.Path]::Combine(
+      $Destination,
+      ([string]$record.path).Replace('/', [System.IO.Path]::DirectorySeparatorChar)
+    )
+    $parent = [System.IO.Path]::GetDirectoryName($target)
+    [void][System.IO.Directory]::CreateDirectory($parent)
     $source = $entry.Open()
     try {
       $output = [System.IO.File]::Open($target, [System.IO.FileMode]::CreateNew, [System.IO.FileAccess]::Write, [System.IO.FileShare]::None)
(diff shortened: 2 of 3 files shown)


Plumvery:
End-to-end result, as promised.

Release build of this branch, installed into a throwaway `BinDir` with a valid hosted-install marker (`managed: true`, `marker: valid`) and a scratch `CTX_DATA_ROOT` with `[search] semantic = true`. Windows 11, `cargo 1.98.0`.

I had to build the branch with the workspace version pinned to `1.1.1` for the test — at `main`'s `1.1.3` the upgrade short-circuits to `up_to_date` and never reaches semantic provisioning. That pin is local to the test only and is not part of this PR.

**The extraction step now completes.** Where the unpatched build dies with the null `drive` binding before creating anything, the staged runtime directory is now populated exactly as expected, including ctx's own install manifest:

```
runtime\onnxruntime\1.27.0\.windows-x64.ctx-upgrade-ua_01a041f2-....new\
    ctx-runtime-install.json
    GIT_COMMIT_ID
    LICENSE
    MICROSOFT_VC_RUNTIME_LICENSE.rtf
    ThirdPartyNotices.txt
    VERSION_NUMBER
    lib\onnxruntime.dll
    lib\msvcp140.dll
    lib\msvcp140_1.dll
    lib\vcruntime140.dll
    lib\vcruntime140_1.dll
```

So the defect in #767 is fixed: `extract_runtime_archive` succeeds, `write_runtime_manifest` runs, and the staged tree is complete.

One thing I want to flag rather than paper over: the run then fails **later**, at `install transaction has an unexpected path count`, after the extraction succeeded. I believe that is an artefact of my synthetic setup — a scratch data root whose semantic asset set is incomplete — rather than anything this change touches, since it occurs downstream of the extraction. But I could not A/B it against the unpatched build, because the unpatched build never gets that far. If you recognise the check and it looks like a real second defect, I am happy to chase it separately with whatever repro you want.

Plumvery:
Correction to my previous comment: the `install transaction has an unexpected path count` failure is **not** an artefact of my scratch setup. It reproduces against a real, live data root, and I now have the mechanism.

This does not affect this PR — it happens strictly after the extraction this PR fixes, and the extraction itself succeeds in both environments. I am recording it here only so the earlier speculation does not mislead you, and I will open a separate issue unless you tell me it is intended behaviour.

The persisted journal (`.ctx.exe.upgrade-install-transaction.json`) after the run holds exactly one entry:

```json
"phase": "prepared",
"paths": [
  { "label": "ONNX Runtime sidecar",
    "staged": "\\?\C:\Users\<u>\.ctx\runtime\onnxruntime\1.27.0\.windows-x64.ctx-upgrade-<id>.new", ... }
]
```

`validate_paths_for_platform` rejects it at `if !(2..=6).contains(&journal.paths.len())`. Relaxing that bound would not help, because the next gate is

```rust
if runtimes.len() > 1 || (!runtimes.is_empty() && binary.is_none()) {
    return Err(anyhow!("install transaction has invalid runtime paths"));
}
```

so a journal carrying the legacy `ONNX Runtime sidecar` label is only publishable together with a `ctx binary` path — i.e. only as part of a version-changing upgrade.

The reason I ran into it is that `ctx upgrade` at the same CLI version is exactly the documented repair path:

> Running `ctx upgrade` again at the same CLI version repairs a missing or hash-mismatched selected semantic asset from the signed catalog without changing the CLI version.

On this machine that repair selects the **legacy** `plan.metadata.onnxruntime` artifact rather than a signed-catalog `Semantic *` asset, and the legacy label cannot be published without a binary replacement. So after this PR lands, a Windows user whose runtime is missing still cannot repair it in place — they need an actual version change to carry it.

I have not tried to fix that here; it looks like a selection/policy question rather than a bug I should guess at. Happy to file it with the journal dump and repro if that is useful.

Plumvery:
Filed the follow-up as #769, as offered above. It is independent of this PR — the extraction fixed here succeeds; #769 is the gate immediately after it.

PR #798, opened 2026-08-28
fix(codex): preserve bounded oversized semantic rows

### LINKED ISSUE #793: Large Codex tool output drops the complete transcript route
ctx 1.2.0 drops a complete Codex transcript route when one `custom_tool_call_output` contains inline `image_url` data and the encoded Core record is larger than 8 MiB. `ctx doctor` reports `JSONL semantic page exceeds the 8388608 byte limit`. The lexical generation then contains no events from that rollout.

To reproduce, create an isolated Codex rollout with `session_meta`, `custom_tool_call`, and `custom_tool_call_output` records. Set the output to an array with text and a 9 MiB base64 data URL. Run `ctx --data-root /tmp/ctx-repro import --provider codex`, then run `ctx --data-root /tmp/ctx-repro list events --provider codex --source-format codex_session_jsonl --content none --limit 1`; the import reports a source failure, and the list returns no events.

The Codex projector accepts content up to 16 MiB, but the shared JSONL publisher limits a semantic page to 8 MiB. Compare the encoded Core record with the shared page limit before publication. If it is too large, retain the event identity and metadata, mark its content as omitted, and publish its siblings normally.

yusoofsh:
## Summary

- keep the shared JSONL semantic page limit at 8 MiB
- split Codex semantic output into bounded pages before publication
- allow only a provider-bounded oversized record to occupy a larger singleton page
- preserve record order and fail if bounded pages remain undrained

## Why

Codex accepts a bounded single projected row larger than the shared 8 MiB JSONL page limit. The semantic executor previously wrapped the scanner output in a generic page, so one valid large `custom_tool_call_output` rejected the complete transcript route.

This change keeps the ordinary 8 MiB limit for generic and multi-record pages. It uses Codex's existing `MAX_CODEX_SOURCE_BACKED_SINGLE_ROW_PAGE_BYTES` contract only for a single record. This preserves the full event instead of omitting valid content.

## Verification

- `cargo test -p ctx-history-jsonl semantic_pages_allow_only_explicit_bounded_singletons_above_default_limit`
- `cargo test -p ctx-history-provider-codex nativepath` (45 passed)

Fixes #793

### DIFF
--- a/crates/ctx-history-jsonl/src/family/route/scanner.rs
+++ b/crates/ctx-history-jsonl/src/family/route/scanner.rs
@@ -197,11 +197,15 @@ impl<E: super::super::JsonlFamilyError> JsonlFamilyOptimizedLeafOutcome<E> {
 #[derive(Debug)]
 pub struct JsonlFamilySemanticPage {
     records: Vec<CoreRecord>,
+    encoded_byte_limit: usize,
 }
 
 impl JsonlFamilySemanticPage {
     pub fn new(records: Vec<CoreRecord>) -> Self {
-        Self { records }
+        Self {
+            records,
+            encoded_byte_limit: PAGE_MAX_BYTES,
+        }
     }
 
     /// Splits projected records into publication pages without changing their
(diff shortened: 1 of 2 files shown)


yusoofsh:
Closing as a duplicate of #796, which already fixes #793. Thanks for the correction.

PR #822, opened 2026-08-29
fix: query ready semantic state without daemon

### LINKED ISSUE #821: Allow daemon-free semantic queries from ready projections during passive refresh
## Summary

The daemon-free semantic contract in #725 is still incomplete after #726 and #772. A ready semantic projection cannot be queried by `ctx search --refresh off` or the default `--refresh background` path when the daemon is disabled.

On ctx 1.2.2, strict semantic search fails with:

```text
semantic_daemon_disabled
local semantic search requires automatic indexing or an available daemon query service
```

Hybrid search falls back to lexical even when the active Core generation has a compatible ready semantic projection.

## Reproduction

1. Configure manual indexing with semantic search enabled and no persistent daemon.
2. Run `ctx search "first query" --backend semantic --refresh wait` to publish Core and reconcile semantic state.
3. Confirm the finite worker exits.
4. Run `ctx search "follow-up" --backend semantic --refresh off`.

Expected: query the ready semantic generation locally, without starting a worker or changing projection state.

Actual: `semantic_daemon_disabled`.

The default background path has the same gap: it remains refresh-inert, but cannot use the ready semantic generation locally.

## Proposed behavior

- In manual mode, `off` and `background` use a read-only foreground query adapter.
- Exact-generation preflight must succeed before model loading or vector search.
- Query embedding may load only verified cached model assets.
- These paths must not download a model, reconcile semantic coverage, or write projection state.
- Missing or stale semantic state must return the existing typed semantic readiness error.
- `wait` keeps its current bounded reconciliation behavior.
- MCP and daemon-backed query behavior remain unchanged.

This restores the `off` and `background` portion of the contract described in #725.

yusoofsh:
Closes #821

## Summary

- add a read-only foreground semantic query mode for daemon-free passive searches
- allow both `--refresh off` and manual `--refresh background` to query an exact ready semantic generation
- load query embedding models from verified cache only on passive paths
- preserve `--refresh wait` reconciliation and existing daemon/MCP behavior
- document the completed daemon-free semantic contract

## Why

#725 specified that daemon-free `off` and `background` searches may query an existing compatible semantic generation locally. #726 implemented the finite `wait` path, while #772 made ready-generation preflight non-blocking. The passive paths still reported `semantic_daemon_disabled` or fell back to lexical.

This change separates foreground execution into two modes:

- `ReadOnly`: exact-generation preflight, cache-only model loading, and vector query; no reconciliation or model acquisition
- `Reconcile`: the existing explicit `--refresh wait` behavior

## Verification

- all nine affected Bazel package suites passed, including 59/59 `ctx-history-index-generation` tests
- strict Clippy passed for all nine affected packages
- repository format, diff, policy, ownership, and LOC gates passed
- manual semantic `--refresh off` and `--refresh background` query paths were exercised without leaving a resident daemon
- the apparent 15 generation-retention failures under direct Cargo were isolated to the invoking shell's permissive umask; the supported Bazel environment and a controlled umask A/B run pass 59/59 with no source change

Native qualification:

- macOS: all nine affected package suites and the reclamation checks passed on the reviewed production tree; the final delta changes only Windows-gated tests, so that evidence remains applicable
- Windows exact head `71259ada20ed608f6e52d3a72dab5700e52d5c29`: sealed source/vendor/payload manifests passed; 8/8 owner/ACL cases, 2/2 pointer replacement/fencing cases, the passive semantic retained-lock case, and the corrected passive-root lock case all passed; final structural verification passed and the process exited 0

The contributor's original commits and authorship remain in the published ancestry; maintainer bridge and correction commits are additive.


yusoofsh:
Verified again against the stock v1.2.3 release: a daemon-free hybrid search with `--refresh off` still degrades to lexical with `semantic_daemon_disabled`, even when the compatible semantic projection is already ready.

I also re-ran the focused PR tests from the current head (`5a64da94`):

```text
ctx-daemon-cli query_adapter::tests: 20 passed; 0 failed
ctx-history-cli manual foreground/passive tests: 2 passed; 0 failed
ctx-history-cli daemon-disabled fallback tests: 2 passed; 0 failed
cargo fmt --all -- --check: passed
```

The first test run used macOS's default temporary directory and hit `SQLITE_CANTOPEN_SYMLINK`. Re-running with `TMPDIR=/private/tmp` passed, so that failure was test-environment-specific rather than a branch failure.

This PR remains the missing piece for the intended daemon-free workflow: use `--refresh wait` once when ctx is triggered to reconcile semantic state, then let `--refresh off` query that ready state without starting a daemon. Review would be appreciated.

PR #824, opened 2026-08-29
fix: preserve native ONNX loader error details

yusoofsh:
## Summary

Surface the native ONNX Runtime loader detail when ctx reports a candidate load failure.

## Why

The pinned ort dynamic loader retains the OS loader message inside its public LoadDynamicError::Dlopen variant, but does not expose that message through its ordinary error source chain. This change extracts that nested libloading cause explicitly, so normal ctx diagnostics retain both the candidate path and the native loader detail.

## Validation

- cargo fmt --all -- --check
- cargo test -p ctx-semantic-model --lib
- cargo clippy -p ctx-semantic-model --all-targets -- -D warnings

The regression test performs a real dynamic load of a nonexistent ONNX library and verifies that its native loader message reaches the diagnostic.

## Non-goals

Diagnostics only; #658 remains open. This does not alter signatures, code signing, library validation, runtime selection, loading behavior, or release assets.

### DIFF
--- a/crates/ctx-semantic-model/src/model_runtime/onnx.rs
+++ b/crates/ctx-semantic-model/src/model_runtime/onnx.rs
@@ -115,6 +115,14 @@ pub(super) struct SemanticOnnxRuntimeCandidate {
     try_even_if_missing: bool,
 }
 
+fn format_runtime_load_failure<E: std::fmt::Display>(
+    source: &str,
+    path: &Path,
+    error: E,
+) -> String {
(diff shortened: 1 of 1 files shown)


luca-ctx:
Superseded by [PR #836](https://github.com/ctxrs/ctx/pull/836), now merged. It preserves the two original commits authored by Yusoof Moh and their contributor attribution. Thank you, @yusoofsh, for the contribution. #658 remains open for release signing remediation.

PR #825, opened 2026-08-29
fix: default foreground Core ML embeddings to CPU

### LINKED ISSUE #823: Default daemon-free foreground Core ML execution to CPU on macOS
## Problem

On macOS, a daemon-free foreground semantic search can stall during implicit Core ML accelerator setup.

With the stock signed v1.2.3 binary, a finite `--refresh wait` search using the default Core ML compute selection remained in Core ML/ANE model compilation and loading for more than eight minutes. The process had to be interrupted.

The same binary and semantic index completed the search in 29.41 seconds when Core ML was explicitly restricted to CPU:

```sh
CTX_INTERNAL_SEMANTIC_BACKEND=coreml \
CTX_SEMANTIC_COREML_NATIVE_COMPUTE=cpu \
ctx search "semantic search" --backend semantic --refresh wait
```

The successful run reported semantic as effective and ready, completed freshness reconciliation, returned a result, and left no ctx process running afterward.

This matters for manual installations that intentionally run without a daemon and only synchronize when `ctx search` is invoked.

## Expected behavior

When no Core ML compute override is set, finite foreground semantic execution should use a reliable compute mode and avoid implicit ANE compilation.

Explicit compute overrides should continue to be honored. The daemon's existing default should not change.

## Proposed scope

- Default only foreground Core ML acquisition to `CPUOnly` when no compute mode was explicitly selected.
- Preserve explicit `all`, `cpuAndNeuralEngine`, `cpuAndGPU`, and `cpuOnly` selections.
- Preserve the daemon's existing compute default.

Related context: #726 recorded a long fresh Core ML/ANE load, while #820 tracks finite workers that survive interruption. This issue is specifically about the implicit compute choice for foreground semantic execution.

yusoofsh:
## Summary

- Default foreground Core ML model acquisition to CPU when no compute mode was explicitly selected.
- Continue honoring every explicit Core ML compute override.
- Leave daemon acquisition behavior unchanged.

## Why

On macOS, the stock v1.2.3 binary remained in implicit Core ML/ANE compilation and model loading for more than eight minutes during a daemon-free `--refresh wait` search. The same binary completed in 29.41 seconds with `CTX_SEMANTIC_COREML_NATIVE_COMPUTE=cpu` and left no background ctx process.

Foreground searches are finite, user-triggered work. A reliable implicit default is more useful there than waiting indefinitely on accelerator setup. Users who want an accelerator can still select one explicitly.

## Validation

```text
cargo fmt --all -- --check
cargo test -p ctx-semantic-model --no-default-features
94 passed; 0 failed

cargo test -p ctx-semantic-model foreground_coreml_defaults_to_cpu_but_honors_explicit_compute_mode -- --nocapture
1 passed; 0 failed
```

The regression test checks that foreground acquisition defaults to CPU and that explicit `all`, `cpuAndNeuralEngine`, `cpuAndGPU`, and `cpuOnly` selections remain unchanged.

Closes #823

### DIFF
--- a/crates/ctx-semantic-model/src/configuration.rs
+++ b/crates/ctx-semantic-model/src/configuration.rs
@@ -131,6 +131,7 @@ pub struct SemanticModelConfig {
     backend_preference: SemanticBackendPreference,
     backend_preference_error: Option<String>,
     coreml_compute_mode: SemanticCoreMlComputeMode,
+    coreml_compute_mode_explicit: bool,
     coreml_compute_mode_error: Option<String>,
     thread_override: Option<usize>,
     batch_size_override: Option<usize>,
@@ -144,6 +145,7 @@ impl SemanticModelConfig {
             backend_preference: SemanticBackendPreference::Auto,
             backend_preference_error: None,
             coreml_compute_mode: SemanticCoreMlComputeMode::All,
+            coreml_compute_mode_explicit: false,
             coreml_compute_mode_error: None,
             thread_override: None,
             batch_size_override: None,
@@ -164,6 +166,7 @@ impl SemanticModelConfig {
 
     pub fn with_coreml_compute_mode(mut self, mode: SemanticCoreMlComputeMode) -> Self {
         self.coreml_compute_mode = mode;
+        self.coreml_compute_mode_explicit = true;
         self.coreml_compute_mode_error = None;
         self
     }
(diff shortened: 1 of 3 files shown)


luca-ctx:
Superseded by and merged as #834. Thank you @yusoofsh for reporting #823 and contributing the original fix. The replacement preserved your authored patch as f3520bd8d5acb2ca376bbc7f596673f0b754a2e7, then added the maintainer scoping correction so the CPU default applies only to foreground composition. Closing this source PR now that the replacement is on `main`.

Issue #928, opened 2026-09-05
ctx could not complete its managed lifecycle handoff


can5162:
PS C:\windows\system32> irm https://ctx.rs/install.ps1 | iex
Installing ctx 1.3.1...
install.ps1: installed ctx could not complete its managed lifecycle handoff; the existing executable was retained
At line:22 char:5
+     throw "install.ps1: $Message"
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (install.ps1: in...le was retained:String) [], RuntimeException
    + FullyQualifiedErrorId : install.ps1: installed ctx could not complete its managed lifecycle handoff; the existin
   g executable was retained

esafak:
Same on MacOS. Existing version is 1.1.1:

```bash
Installing ctx 1.3.1.. error: installed ctx could not complete its managed lifecycle handoff (status 1); the existing executable was retained
Installing ctx 1.3.1...
```

Possibly relevant:

```bash
> ctx upgrade
✗ release 1.3.1 is newer than this ctx build but has no complete ONNX Runtime sidecar metadata; refusing a
  downgrade-compatible legacy update
```


Is there a way to get verbose output?
Also, I wish there was a lightweight way for ctx to surface new versions (e.g., in the first query in a session) and indexing problems.

Iritscen:
For what it's worth, a fresh install of 1.1.3 on my Mac also failed.

```
Installing ctx 1.3.1...ctx child output:
Installing ctx 1.3.1.  error: unexpected argument '--ctx-core-managed-pair-apply-v1' found
```

Not sure this is the same issue though.

PR #952, opened 2026-09-08
Report Semantic startup failures and document runtime installation

### LINKED ISSUE #948: Semantic enable reports preparing when runtime is missing
On Linux x86_64, enabling semantic search after a direct GitHub release installation reports that search is preparing, but background model loading fails because ONNX Runtime is missing. The supported way to provision CUDA for this installation is unclear.

### Environment

- ctx 1.3.1, installed from the GitHub release's Linux x64 Core binary.
- NVIDIA GPU visible to `nvidia-smi`.
- Built-in semantic executor, automatic indexing enabled.
- Direct installation was used after the hosted installer failed as reported in #947.

### Observed behavior

After running:

```sh
ctx setup
ctx semantic enable
ctx semantic status
```

The human-facing output says:

```text
Semantic search is preparing
Model acquisition or semantic indexing is still in progress.
```

Once initial history indexing reached model loading, detailed daemon status showed:

```text
last_run_reason: model_load_failed
last_run_status: skipped
last_error: no ONNX Runtime dynamic library candidates were found for linux-x64; set an absolute path with CTX_ONNXRUNTIME_DYLIB, ORT_DYLIB_PATH, CTX_ONNXRUNTIME_DIR, CTX_ONNXRUNTIME_CACHE_DIR, or CTX_RUNTIME_DIR
```

The high-level semantic status continued to report pending/preparing. This makes an actionable missing dependency look like ordinary background progress.

### CUDA provisioning evidence

The v1.3.1 source expects ONNX Runtime 1.27.0 and the CUDA asset `ctx-onnxruntime-linux-x64-cuda12.tar.zst`. The accelerator loader validates a canonical runtime layout, an installation manifest, and bundled dependencies:

- [Runtime version and CUDA asset selection](https://github.com/ctxrs/ctx/blob/v1.3.1/crates/ctx-semantic-model/src/model_runtime/onnx.rs)
- [Accelerator runtime validation](https://github.com/ctxrs/ctx/blob/v1.3.1/crates/ctx-semantic-model/src/model_runtime/onnx/verified_runtime.rs)

During investigation:

- The GitHub v1.3.1 release assets contained Core executables and `SHA256SUMS`, but no runtime sidecars.
- Stable metadata at `https://cli.ctx.rs/functions/v1/releases/stable/ctx-release-metadata.env` contained no `CTX_RELEASE_ONNXRUNTIME_` or `CTX_RELEASE_SEMANTIC_` fields.
- Requesting `ctx-onnxruntime-linux-x64-cuda12.tar.zst` under that metadata's artifact base, `https://cli.ctx.rs/storage/v1/object/public/releases/artifacts/stable/1.3.1/`, returned HTTP 404. This establishes absence at that location, not that the asset is unavailable everywhere.
- `ctx upgrade --dry-run --format json` rejected the unmanaged installation and directed it back to the hosted installer.

No CUDA inference failure has been observed: model loading is blocked before successful inference.

### Narrowed cause and loader contract

The observed failure is a runtime-provisioning failure before inference: the Core executable and local model files are present, but the required native runtime is absent. The upstream release-publication reason for that absence has not been established.

For CUDA, the source shows that supplying an arbitrary ONNX Runtime library path is not sufficient:

- `verified_accelerator_runtime_candidates` searches runtime roots for `onnxruntime/1.27.0/linux-x64-cuda12/lib/libonnxruntime.so`. Unlike the general CPU candidate search, it does not consume the direct `CTX_ONNXRUNTIME_DYLIB` or `ORT_DYLIB_PATH` selections.
- `validate_runtime_candidate` requires `ctx-runtime-install.json` with the expected version/platform, `manager = "ctx-hosted-installer"`, and `metadata_trust = "signed-release-metadata"`.
- It also verifies the exact file allowlist, sizes, and hashes, including the bundled CUDA/cuDNN dependencies.

These checks are in [verified_runtime.rs](https://github.com/ctxrs/ctx/blob/v1.3.1/crates/ctx-semantic-model/src/model_runtime/onnx/verified_runtime.rs). The distinction from general runtime-path lookup is visible in [onnx.rs](https://github.com/ctxrs/ctx/blob/v1.3.1/crates/ctx-semantic-model/src/model_runtime/onnx.rs).

Consequently, CUDA recovery needs a supported way to provision the complete expected runtime bundle and manifest. The generic missing-library advice about setting a direct dynamic-library path does not explain this CUDA-specific requirement. This is a provisioning/documentation gap in addition to the high-level status hiding `model_load_failed`; it is not evidence of a CUDA kernel or GPU failure.

### Expected behavior

- Surface the missing-runtime/model-load error in human-facing semantic status instead of only reporting preparing.
- Document a supported way to install the CUDA runtime and its required layout for direct-release installations, or clearly state if this requires a hosted installation.
- If automatic runtime acquisition is intended, make the necessary runtime artifacts discoverable and report why acquisition cannot proceed.

This is related to #947, but the missing-runtime reporting and direct-install provisioning path are separate from the installer's rejected internal command.

camjac251:
Fixes #948. Release metadata and older-install verification remain tracked in #953 and #928.

## Summary

Surface persisted Semantic startup failures instead of reporting that the runtime is still preparing. The shared daemon report now preserves the failure reason and error detail, and status, doctor, and Semantic status project those facts consistently. Keyword-search readiness and the existing Semantic projection inventory remain intact.

Retryable startup failures are reported as failures while retaining their existing retry classification and backoff. Resource-pressure deferrals remain distinct, and pending configuration changes take precedence over an older failure.

Document the hosted managed installation route for the complete CUDA runtime bundle, including the distinction from direct CPU library overrides. This integration retains the contributor's bounded error rendering and tests; it does not add manual runtime-install commands, a new trust tier, or model/cache provisioning changes.

## Verification

- Independently reproduced the persisted missing-runtime failure being reported as pending, then verified the corrected shared report.
- Passed all 212 canonical CI targets, including actual missing-library and model-acquisition failure checks and retry/backoff assertions.
- Verified configuration-transition and disabled-state precedence, unchanged lexical readiness, and human-facing error rendering.
- Documentation and affected checks cover the final installation guidance and JSON contract clarification.

This change fixes failure reporting and documents the supported provisioning route; it does not claim a new CUDA inference qualification or publish a client release.


camjac251:
@luca-ctx any possibility this could make it in

